PR #4, opened 2024-03-14
Update ebook2audiobook.py to fix blank fragment

Rihcus:
So I found the sent_tokenize() function can occasionally give a blank output (I'm not sure why)

sentences = sent_tokenize(chapter_text, language='italian' if language == 'it' else 'english')

This causes issues around the loop in line 419 since if a blank fragment is put line 424 tts.tts_to_file will crash since no text is present to convert.

Solution (I think):

add a if fragment != "": under for fragment in fragments:

Disclamer I have very minimal python knowledge and this might not be the most elegant solution.


DrewThomasson (commented):
nice, looks like it should work, merging.

PR #49, opened 2024-11-18
Update colab_ebook2audiobookxtts.ipynb

pafend:
Add "punkt_tab" to nklt call to be able to process more ebooks


DrewThomasson:
Looks good 👍

PR #76, opened 2024-12-12
added updated morgan freeman model to list

DrewThomasson:
added my updated re-trained morgan freeman model to the list of fine-tuned xtts models with the link to its respective ref.wav file to use in inference of this model.

### DIFF
--- a/lib/conf.py
+++ b/lib/conf.py
@@ -60,10 +60,17 @@ ebook_formats = ['.epub', '.mobi', '.azw3', 'fb2', 'lrf', 'rb', 'snb', 'tcr', '.
 audiobook_format = 'm4b' # or 'mp3'
 audio_proc_format = 'wav' # only 'wav' is valid for now
 
-xtts_fine_tuned_voice_actors = {
+fine_tuned_voice_actor_models = {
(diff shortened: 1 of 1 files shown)


DrewThomasson:
@ROBERT-MCDOWELL 

Do you approve this merge?

ROBERT-MCDOWELL:
how do you want to use them in production mode? don't forget that now we support more than 1100 languages and we should be aware of how to list and order them....

DrewThomasson:
hm.... not sure

i know these are just the xtts fine-tuned models but i suppose the ref.wav files with them could be used for other voice cloning models too...

DrewThomasson:
but these are english models and english voices.....

DrewThomasson:
hm.... we should probs call on discord about his or something so we can communicate what eachothers plans and goals are

DrewThomasson:
I'll look at drawing up some interface design ideas and just give me any times your open to call so we can shoot ideas back and forth at eachother

DrewThomasson:
@ROBERT-MCDOWELL 

what do you think of that layout?

I changed the David attenburo model details to what i think you meant

PR #80, opened 2024-12-14
added voice actor voices

DrewThomasson:


DrewThomasson:
@ROBERT-MCDOWELL 

looks good?

Added the voice audio files in the manner that I think we agreed on

ROBERT-MCDOWELL:
LGTM, the name of the wav file must be the same of the name of the model in conf.py -> models (example in discord comments)

DrewThomasson:
Yeah it's the name of the model folder from the huggingface

PR #85, opened 2024-12-15
Made a fix for loading and downloading the fine-tuned tts models

DrewThomasson:
We don't have to use this but it'll at least help you get a better idea on what to do to fix the issue that we were having.

### DIFF
--- a/lib/conf.py
+++ b/lib/conf.py
@@ -53,14 +53,19 @@ models = {
             "lang": "multi",
             "folder": "tts_models--multilingual--multi-dataset--xtts_v2",
             "api": "tts_models/multilingual/multi-dataset/xtts_v2",
-            "voice": default_voice_file
+            "voice": default_voice_file,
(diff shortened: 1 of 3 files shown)


DrewThomasson:
@ROBERT-MCDOWELL 

:)?

DrewThomasson:
Cause I realize the only issue we are having at this point seems to be that it isn't downloading the fine-tuned model

ROBERT-MCDOWELL:
we don't need to download it manually, the fact to call TTS(/hf/path) should do it. it worked before so there is no reason it does not work anymore unless a little glitch is the code that perturbs the cache creation.
downloading manually creates things more complex like check everytime if the folder exists etc... and ask more maintenance for long term code.

DrewThomasson:
Mmm fair fair still learned some lol

ROBERT-MCDOWELL:
I'm working on it. in same time a server update is causing a segfault of a lot of compiled app I'm running. I must check if it's a major issue from glibc or my old apps which are not compilable anymore... a nightmare...

PR #88, opened 2024-12-17
added  ref.wav in custom model upload

ROBERT-MCDOWELL:
custom_model still has an issue. zip cannot be extracted. I think it's a inattention mistake.... will see later


DrewThomasson:
👍

PR #107, opened 2024-12-25
Fixed f-string syntax errors :)

DrewThomasson:
@ROBERT-MCDOWELL 

?

:)

### DIFF
--- a/app.py
+++ b/app.py
@@ -187,7 +187,7 @@ Linux/Mac:
                 if os.path.exists(args['ebooks_dir']):
                     new_ebooks_dir = os.path.abspath(args['ebooks_dir'])
                 else:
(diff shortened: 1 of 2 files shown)


DrewThomasson:
Thats needed to allow it run run in the hugginface space and the docker is why lol

DrewThomasson:
Just making sure that your also approving every merge I do so we don't run into any errors on your end lol 

:)

ROBERT-MCDOWELL:
yeah I can do it only after you merged my PR, if not we will have the same shite that last time ;-|

DrewThomasson:
Kk well I'll let u fix it then lol

Don't want conflicts in merges lol

PR #113, opened 2024-12-25
V2.0 glossary test

DrewThomasson:
@ROBERT-MCDOWELL 

Was looking at adding a Glossary to the README 🤔 

Only Downside is for it to function we have to remove the emojis from the chapter titles :(

Thoughts?

Or any suggestions on where to put it if you approve of it? 🤔


ROBERT-MCDOWELL:
README can survive without emojis I think :)

DrewThomasson:
tru tru 👍 

So your fine with that placement of it?

ROBERT-MCDOWELL:
LGTM, language code must be 3 letters btw ;)

DrewThomasson:
Oo good catch

PR #133, opened 2024-12-27
Regex import error fix: import re instead of import regex as re

### LINKED ISSUE #135: ModuleNotFoundError: No module named 'regex'
When running the bash script this error occurs:
```
  File "/home/leon/code/ebook2audiobook/app.py", line 3, in <module>
    import regex as re
ModuleNotFoundError: No module named 'regex'
```
#133 pull request fixes it by just importing re instead of regex.

If specifically the regex library is necessary for functionality, then the pull request #134 is helpful.
It adds it to requirements.txt where it was missing in the current version of the code.

Leon-Sander:
Running the script throws this error:
  File "/home/leon/code/ebook2audiobook/app.py", line 3, in <module>
    import regex as re
ModuleNotFoundError: No module named 'regex'

This PR fixes this error by setting the import to `import re` instead of `import regex as re`

Fixes #135

### DIFF
--- a/app.py
+++ b/app.py
@@ -1,6 +1,6 @@
 import argparse
 import os
-import regex as re
+import re
 import socket
 import subprocess
 import sys


ROBERT-MCDOWELL:
regex is necessary as re package does not manage with full UTF8 and UTF16.

PR #134, opened 2024-12-27
missing regex in requirements added

### LINKED ISSUE #135: ModuleNotFoundError: No module named 'regex'

Leon-Sander:
Running the script throws this error:
File "/home/leon/code/ebook2audiobook/app.py", line 3, in
import regex as re
ModuleNotFoundError: No module named 'regex'

This PR adds regex to the requirements.txt

Fixes #135

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,6 +16,7 @@ mecab-python3
 pydub
 pypinyin
 ray
+regex
 transformers
 translate
 tqdm


ROBERT-MCDOWELL:
I guess you run the docker? seems that the docker file still has an old pip pacckage list. I just updated it for the next git update
regex is necessary as re package doesn't manage UTF8 and UTF16 chars correctly

Leon-Sander:
But why declining the PR if it fixed it?

ROBERT-MCDOWELL:
no it didn't, regex must be installed, not re.

PR #157, opened 2024-12-29
GitHub Actions: Lint Python code with Ruff

cclauss:
https://docs.astral.sh/ruff is an extremely fast Python linter and code formatter, written in Rust. With [over 800 linting rules](https://docs.astral.sh/ruff/rules), ruff can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [Black](https://github.com/psf/black), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [pyupgrade](https://pypi.org/project/pyupgrade/), [autoflake](https://pypi.org/project/autoflake/), and more, all while executing tens or hundreds of times faster than any individual tool.

### DIFF
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -1,0 +1,16 @@
+# https://docs.astral.sh/ruff/linter
+name: ruff
+on:
+  push:
+    branches: 
+      - 'main'
+  pull_request:
+    branches: 
+      - 'main'
+jobs:
+  ruff:
+    name: ruff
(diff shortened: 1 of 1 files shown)


ROBERT-MCDOWELL:
could you please open a discussion rather than promote your for sure interersting information rather than creae a PR that has nothing to do to fix bugs or add a new feature on our repo? it will be more respectful and less intrusive, thanks.

PR #172, opened 2025-01-01
Added details about how to build and making the xtts model baked in d…

DrewThomasson:
…uring the build

Added details about how to build and making the xtts model baked in during the build...


ROBERT-MCDOWELL:
no python 3.12? seems to be faster and less lib conflicts...

DrewThomasson:
Hm good point I'll see if that works

DrewThomasson:
@ROBERT-MCDOWELL 

kk updated to python 3.12 now and it seems to work! :)

PR #201, opened 2025-01-06
Fix conda install links for different architectures

Daniil-Gusev:


ROBERT-MCDOWELL:
not ok?

Daniil-Gusev:
sorry for the slightly messy commit history, i forgot to remove the debug info and then got a bit confused with using git --amend :(

ROBERT-MCDOWELL:
the 2 last lines must be removed

Daniil-Gusev:
> the 2 last lines must be removed

I removed the last two lines, which were debug lines. Or do you mean that it should be possible to run eb2ab on apple silicon processors using Rosetta? Which conda script should be installed in that case, x86_64?

ROBERT-MCDOWELL:
I was talking about the debug lines. all fine now. THANKS!

PR #206, opened 2025-01-08
Test script for testing every fairseq language

DrewThomasson:
will run each in headless mode in a docker one by one

and give log for each language code organized by success and failed attempts


ROBERT-MCDOWELL:
would it be worth to run all its tts  compatible language? like eng will run xtts and fairseq..

DrewThomasson:
;)

PR #243, opened 2025-01-22
Update README.md to add dependency install instructions

HydeZero:
When trying to run this app on Zorin OS 17.2 core, I found that `mecab` failed to install due to a missing mecab install. I resolved it by installing mecab. To prevent this issue from happening to other people, I took the commands from colab and condensed them down so I could add them to README.md. Feel free to move the dependency install instructions around if you find they don't quite fit into the location I put them in (Running Gradio Web Interface).

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,10 +125,19 @@ git clone https://github.com/DrewThomasson/ebook2audiobook.git
 
 Specify the language code when running the script in  mode.
 
-
 ### Launching Gradio Web Interface
-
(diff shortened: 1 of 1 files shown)


ROBERT-MCDOWELL:
this issue has been reported many times and it's on the next version already.... thanks anyway

HydeZero:
You're welcome. Thanks for letting me know 👍

PR #275, opened 2025-02-15
Updated Ubuntu Workflow file, Added a custom xtts model test

DrewThomasson:
@ROBERT-MCDOWELL 

Does the command syntax look right to you? 🤔

PR #277, opened 2025-02-15
Added workflow testing for loading custom xtts zip files

DrewThomasson:
Updated to main but with new custom xtts model form zip loading test in single and batch added to workflow

### DIFF
--- a/.github/workflows/run-ubuntu-headless-ebook2audiobook.yml
+++ b/.github/workflows/run-ubuntu-headless-ebook2audiobook.yml
@@ -51,6 +51,7 @@ jobs:
           echo "This is test sentence 3 ." > workflow-testing/test3.txt
           echo "This is test sentence 4 ." > workflow-testing/test4.txt
           echo "This is test sentence 5 ." > workflow-testing/test5.txt
+
       - name: Default headless single test python app.py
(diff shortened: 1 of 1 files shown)


DrewThomasson:
@ROBERT-MCDOWELL 

seeing how the workflow tests on the updates you made to the main that should of fixed the [issue](https://github.com/DrewThomasson/ebook2audiobook/pull/275)

DrewThomasson:
Hope the new workflow tests run without any issues! 🤞

DrewThomasson:
@ROBERT-MCDOWELL 

heck runs into an issue, 

But I think the workflow is running correctly tho

So I think i'll still merge it into the main?

Then you can see the latest bug when it tries to load a custom xtts model .zip

DrewThomasson:
Merging as its confirmed that this is testing correctly, its giving the same errors given when this guy was doing the same kind of test 

https://github.com/DrewThomasson/ebook2audiobook/issues/278

ROBERT-MCDOWELL:
> @ROBERT-MCDOWELL
> 
> heck runs into an issue,
> 
> But I think the workflow is running correctly tho
> 
> So I think i'll still merge it into the main?
> 
> Then you can see the latest bug when it tries to load a custom xtts model .zip

ok you can merge

PR #286, opened 2025-02-16
Made Dockerfile build full image with xtts built in

DrewThomasson:


DrewThomasson:
@ROBERT-MCDOWELL 

just alerting you too this change 👍

PR #298, opened 2025-02-17
Clone repo without any history to save storage in docker image

DrewThomasson:
Clone repo without any history or .git to save storage in docker image


DrewThomasson:
Cause I don't think people are going to be using the git history in the docker image?

Whats your thoughts @ROBERT-MCDOWELL 

👍 ?

ROBERT-MCDOWELL (approved):
absolutely!, I'm even sure many other stuff can be removed too.

DrewThomasson:
Nice! 👍

PR #294, opened 2025-02-17
fixed docker run commands in readme

DrewThomasson:
now they have --script_mode full_docker added 

so it now uses 

python app.py --script_mode full_docker

PR #299, opened 2025-02-17
Swapped CMD with ENTRYPOINT

DrewThomasson:
Should make it easier for users as it will append to the end of it instead of replacing lol

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,4 +55,4 @@ RUN rm test.txt
 EXPOSE 7860
 
 # Start the Gradio app with the required flag
-CMD ["python", "app.py", "--script_mode", "full_docker"]
+ENTRYPOINT ["python", "app.py", "--script_mode", "full_docker"]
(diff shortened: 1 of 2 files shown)


DrewThomasson:
@ROBERT-MCDOWELL 

And then I will update the readme to align with those changes here to make it cleaner :)

at the PR https://github.com/DrewThomasson/ebook2audiobook/pull/294

DrewThomasson:
Now the commands will easily append to the end of the docker run commands

Makes the command more pretty and elegant 

:)

@ROBERT-MCDOWELL 
👍 ?

ROBERT-MCDOWELL (approved):
didn't know you can use another than CMD. ENTRYPOINT is different behavior?

/my/path/to/file rather than /input_folder/YOUR_INPUT_FILE.TXT
newbies will think maybe txt is mandatory ;)

DrewThomasson:
Yup

CMD will get overwritten

ENTRYPOINT will be appended

DrewThomasson:
kk changed it to 

 /input_folder/YOUR_EBOOK_FILE
 
 https://github.com/DrewThomasson/ebook2audiobook/pull/294

PR #300, opened 2025-02-17
Fixing how the version tag name is grabbed

DrewThomasson:


DrewThomasson:
@ROBERT-MCDOWELL 

Just alerting you to this I think this should fix the issue I was having

ROBERT-MCDOWELL:
kk :)

DrewThomasson:
Confirming that this did indeed fix the issue after testing on my end

:)

PR #301, opened 2025-02-17
--platform=linux/amd64 removed n docker run command

DrewThomasson:
--platform=linux/amd64 is not needed in docker run command

Because of this I removed it, further shortening the docker run command

:)

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,11 +305,11 @@ To run the Docker container and start the Gradio interface, use the following co
 
  -Run with CPU only
 ```powershell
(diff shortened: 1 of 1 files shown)


DrewThomasson:
@ROBERT-MCDOWELL 
👍 ?

ROBERT-MCDOWELL (approved):
yup, good job!

DrewThomasson:
Also Removed the `-it` from the docker run commands

As the user is never interacting with the docker shell anyway

so shortened even more now!

:)

DrewThomasson:
kk merging now then with that modification

I'll update the other languages readme commands later

PR #329, opened 2025-02-19
Added fine-tuned model test

DrewThomasson:
For auto-testing of the 

`--fine_tuned` parameter


DrewThomasson:
@ROBERT-MCDOWELL 
👍 ?

DrewThomasson:
Made the Fairseq and Xtts tests separated into two different workflow files to save time :)

DrewThomasson:
And just added GitHub Workflows for `Bark`, `Vits`, and `Yourtts`

:)

DrewThomasson:
Ok so the full-run on does all of the tests, that has been set to be the only one that runs automatically with the single  `act` command

This is primarily to easily run locally with a single `act` command 

cause cacheing apparent doesn't work for act locality >:|


And the others are still there for faster more specific runs that can be manually run or run

DrewThomasson:
@ROBERT-MCDOWELL 

Alerting you that I'm merging this to see if it works with showing up as multiple in the actions button

DrewThomasson:
hm

well only shows as one but whatever still should be faster cause they can all be run concurrently I guess....

PR #374, opened 2025-02-25
Memory optimizations for full-test-ubuntu-headless.yml

DrewThomasson:


DrewThomasson:
That should fix any memory issues for the workflow

But I'll also try testing it out locally with `act` to make sure it runs

ROBERT-MCDOWELL (approved):
LGTM

Issue #331, opened 2025-02-19
MPS on Docker


souzadevinicius:
Hello there.
Is it possible to use mps on Docker?
Regards

ROBERT-MCDOWELL:
yes, but for now MPS seems slower than CPU. coqui-tts is not (yet) optmized for it and it needs the devs to patch it with the corresponding class/functions managing MPS correctly

DrewThomasson:
Also we are unable to install `coqui-tts` on the Arm Debian image as we run into rust issues.

So... yeah

The docker image is only available as a `x86` image rn

ROBERT-MCDOWELL:
rustc and cargo must be installed on arm debian, like ubuntu, it's in the .sh script. what errors you can see?

DrewThomasson:
I'll try checking when I find time

But I think I ran into issues when trying to install rustc even

ROBERT-MCDOWELL:
is it only on arm?

DrewThomasson:
> is it only on arm?

Only on the Arm Docker python12 image 

here its this error on only the Arm Docker python12 Image 

```powershelgl
      cargo rustc --lib --message-format=json-render-diagnostics --manifest-path Cargo.toml --release -v --features pyo3/extension-module -- --crate-type cdylib
      error: package `fancy-regex v0.13.0` cannot be built because it requires rustc 1.66 or newer, while the currently active rustc version is 1.63.0
      Either upgrade to rustc 1.66 or newer, or use
      cargo update -p fancy-regex@0.13.0 --precise ver
      where `ver` is the latest version of `fancy-regex` supporting rustc 1.63.0
      error: `cargo rustc --lib --message-format=json-render-diagnostics --manifest-path Cargo.toml --release -v --features pyo3/extension-module -- --crate-type cdylib` failed with code 101
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for sudachipy
Successfully built encodec gruut monotonic-alignment-search docopt gruut-ipa gruut_lang_de gruut_lang_en gruut_lang_es gruut_lang_fr
Failed to build sudachipy

[notice] A new release of pip is available: 24.3.1 -> 25.0.1
[notice] To update, run: pip install --upgrade pip
ERROR: ERROR: Failed to build installable wheels for some pyproject.toml based projects (sudachipy)
root@ba392cfb646c:/# 

```

#Full log: 

[...]

PR #413, opened 2025-03-04
Trying to make it able to build on Arm docker

DrewThomasson:


DrewThomasson:
Note to self:

Looking at this more 

If the rust runs into permission errors somehow 

Then 

I think I'll just create a dockefile for the huggingface space and use the current implantation in main for that

Cause that at needs users and such which is annoying 


And then update the main one to only install and run in root

Then I won't have to deal with that stuff with acess to the rust or whatnot

DrewThomasson:
This is a log of the information I am gathering for making the docker be able to build on ARM natively as well

Partly related to things mentioned here https://github.com/DrewThomasson/ebook2audiobook/issues/331

## New findings:

Just found a PR in coqui-tts where they got it to build on arm in their docker and how they fixed the rust, to build sudachipy on arm

PR being mentioned here: -> 
https://github.com/idiap/coqui-ai-TTS/pull/159

DrewThomasson:
Somehow this reduced the docker image on arm at least by around 5gb????

so the docker image uncompressed size went from 15gb -> 10gb?????????????

What is this magic?????

DrewThomasson:
We got dockerx in the workflow again babyyyyyyyyyyy, yessss

ROBERT-MCDOWELL (approved):
Looks GOOD

DrewThomasson:
:DDD

PR #415, opened 2025-03-05
Only uses self-hosted now full-test-ubuntu-headless.yml

DrewThomasson:


DrewThomasson:
@ROBERT-MCDOWELL 

Just making this only use self-hosted 

idk makes it go to my free hosted faster

the free GitHub won't run the full thing anyway so 🤷

DrewThomasson:
Merging lol

PR #416, opened 2025-03-05
Adding missing installs to workflow (ON hold currently testing...)

DrewThomasson:
Idk trying to make it have everything that the working dockerfile includes

Hopefully it should work now with these additions

if not then next I'll try making each python app.py command run in full docker mode or something

like this does

```workflow
# Do a test run to make sure that the base models are pre-downloaded and baked into the image
RUN echo "This is a test sentence." > test.txt 
RUN python app.py --headless --ebook test.txt --script_mode full_docker
RUN rm test.txt
```

from this dockerfile

https://github.com/DrewThomasson/ebook2audiobook/blob/v25/Dockerfile

### DIFF
--- a/.github/workflows/full-test-ubuntu-headless.yml
+++ b/.github/workflows/full-test-ubuntu-headless.yml
@@ -13,6 +13,7 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
+      
       - name: Set up Python 3.12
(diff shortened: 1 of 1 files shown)


DrewThomasson:
I need to figure out what base image the docker build uses

And see which workflow docker thing is the closest to that idk

Perhaps that'll work

PR #432, opened 2025-03-06
Added DockerSelfTest.sh and test txt files

DrewThomasson:


DrewThomasson:
not needed apparently...

PR #433, opened 2025-03-07
(Will delete when done) Update docker_build_workflow.yml

DrewThomasson:


DrewThomasson:
IF this one passed im adding this to the main under a new PR 

and also will be amking the regular docker build no longer auomtatica as this one does its job automatically

DrewThomasson:
Yup works I think 👍

PR #434, opened 2025-03-07
(Will delete when done) Working self hosted workflow patch 1

DrewThomasson:
Not needed apparently...


DrewThomasson:
Mpstly Just a bunch of scratch paper for me to work with

DrewThomasson:
Yup This one works 👍

PR #447, opened 2025-03-08
fix lite workflow

DrewThomasson:


DrewThomasson:
@ROBERT-MCDOWELL 

Altering you of this 👍

PR #454, opened 2025-03-09
Workflow patch 1

DrewThomasson:
Initial build now only runs on my ARM mac's, they build dockerx faster is why


DrewThomasson:
@ROBERT-MCDOWELL 

Alerting you to this 👍

PR #455, opened 2025-03-09
Workflow patch 1

DrewThomasson:


DrewThomasson:
This way the dev docker test and the docker test that Build-Tests-and pushes the latest that auto-runs in the main Won't come in conflict with eachother

The build test push workflow names its image to test `:dev`

The pure build+Test for the dev side names its image to test `:dev_25`

DrewThomasson:
Alterting you of this change @ROBERT-MCDOWELL

PR #461, opened 2025-03-10
Update README.md

DrewThomasson:
This badge should show up in the readme with a real status....

once this workflow has run once I think?


DrewThomasson:
@Robert-Adding it cause it should work....

op Gota add a `docker builder prune`  thing to the workflows

PR #467, opened 2025-03-10
Removed release trigger from Dev test

DrewThomasson:
Cause that's for the main Build+Test to do

This is redundancy for no reason if I have this also run lol


DrewThomasson:
@ROBERT-MCDOWELL 

Altering you to this

PR #482, opened 2025-03-14
Working Multi-stage Dockerfile

DrewThomasson:


DrewThomasson:
The only thing I have left I think is to Make it not Attempt to build the `xpu` variant on `ARM` as it appears there is no ARM variant of it, soooo I guess only x86 for now for that

DrewThomasson:
You also need to find out of the building the `xpu` variant even works at all on `x86` either, Cause like, You know know it doesn't build on `ARM` And that's it really with your knowledge on it lol

DrewThomasson:
@ROBERT-MCDOWELL 

merging cause I think this works

PR #505, opened 2025-03-17
Making lite test manual again

DrewThomasson:


DrewThomasson:
@ROBERT-MCDOWELL 

IDK why but making this automatic stopped the full test from automatically running so...

Alerting you of this change 👍

PR #556, opened 2025-03-19
V25

DrewThomasson:


DrewThomasson:
Heck, I forgot about the custom-voice bark headless test

DrewThomasson:
I can still force it to build and ignore all other tests Tho if we have to, thats a parameter I added for the docker build+test thing when manually activated

PR #576, opened 2025-03-20
Fixing full-rebuild in lite workflow

DrewThomasson:


DrewThomasson:
@ROBERT-MCDOWELL 
Merging this fix, Now the force factory rebuild option will work again

PR #587, opened 2025-03-21
Removing bark from tests

DrewThomasson:


DrewThomasson:
@ROBERT-MCDOWELL 
Alerting you of this 👍

ROBERT-MCDOWELL:
if you just let a bark test for manual test on v25 it would be great

PR #592, opened 2025-03-21
fixing lite auto test

DrewThomasson:


DrewThomasson:
hm....

Something weird is going on...

I'lll have to fix this later

PR #601, opened 2025-03-21
Fixing torch version variant issues

DrewThomasson:


DrewThomasson:
This is to try to have it dynamically use the latest torch version for the variants if a specific version number isn't found.

But if there is a version found it'll try to use that specific torch version for the variant builds

DrewThomasson:
@ROBERT-MCDOWELL 
Alerting you of this

It should fix the variant build issues,

DrewThomasson:
merging

PR #602, opened 2025-03-21
reverting back to bark as default test....

DrewThomasson:


DrewThomasson:
@ROBERT-MCDOWELL 

idk seeing if just reverting fixes it

PR #614, opened 2025-03-22
Adding pull-always to compose

DrewThomasson:


DrewThomasson:
@ROBERT-MCDOWELL 👍 

Now if they use the compose it will always try to get the latest image

PR #617, opened 2025-03-22
Add pull always to docker run commands

DrewThomasson:


DrewThomasson:
@ROBERT-MCDOWELL 
This way anyone who uses the docker run commands will always be using the latest docker image

PR #618, opened 2025-03-22
Add Pull always to docker run commands in readme

DrewThomasson:


DrewThomasson:
@ROBERT-MCDOWELL 👍 

If anyone says that docker is taking up too much space from this then you can just tell them to prune all dangling images. 

```bash
docker image prune
```

Issue #631, opened 2025-03-26
a chinese epub file will cause this exception


williamxhero:
Error extracting main content pages: string index out of range
Traceback (most recent call last):
  File "C:\Users\admin\Documents\PERSONAL\projs\ebook2audiobook\lib\functions.py", line 605, in get_chapters
    doc_cache[doc] = filter_chapter(doc, session['language'], session['language_iso1'], session['tts_engine'])
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\admin\Documents\PERSONAL\projs\ebook2audiobook\lib\functions.py", line 530, in filter_chapter
    chapter_sentences = get_sentences(text, lang)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\admin\Documents\PERSONAL\projs\ebook2audiobook\lib\functions.py", line 713, in get_sentences
    sentences.extend(split_sentence(sentence.strip()))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\admin\Documents\PERSONAL\projs\ebook2audiobook\lib\functions.py", line 652, in split_sentence
    if sentence[-1].isalpha():
       ~~~~~~~~^^^^
IndexError: string index out of range
Caught DependencyError: Error extracting main content pages: string index out of range
get_chapters() failed!

DrewThomasson:
## Further details in testing:

Confirmed to occur with this short Chinese test txt file -> [chinese_test.txt](https://github.com/user-attachments/files/19469649/chinese_test.txt)

Error `get_chapters() failed!` occurs with the models: 
`xtts` -> `Chinese` ->  [XTTSlog.txt](https://github.com/user-attachments/files/19469792/log.txt)

&

 `fairseq` ->  `Chinese, Min Nan - 閩南語` ->  [Min_Nan_log.txt](https://github.com/user-attachments/files/19469777/log.txt)


Error does not occur with `fairseq` -> `Chinese, Hakka - 客家話` But the output is random noise as seen here -> https://github.com/user-attachments/assets/ee6800b6-048e-4235-ac35-b64c14e132e0

TrojanFish:
same issue

DrewThomasson:
While we work on the fix,

[...]

PR #646, opened 2025-03-28
Fix IndexError in split_sentence when handling empty strings (frequent in Chinese EPUBs)

gaivrt:
This PR fixes an IndexError in `split_sentence()` when the input sentence is an empty string. This issue occurs frequently in Chinese EPUB books due to how some chapters are structured or extracted.

Adding a simple `if not sentence: return []` guard avoids the crash without changing existing behavior.

### DIFF
--- a/lib/functions.py
+++ b/lib/functions.py
@@ -649,6 +649,8 @@ def get_sentences(text, lang):
     def split_sentence(sentence):
         end = ''
         if len(sentence) <= max_chars:
+            if not sentence:
+                return []
             if sentence[-1].isalpha():
                 end = '–'
             return [sentence + end]


DrewThomasson:
@ROBERT-MCDOWELL Is currently assigned to `functions.py` at the moment. 

So we'll have to wait for him to get back in like a week or so for this to be merged...

DrewThomasson:
In the meantime I'll let you promote your fork fix in the -> https://github.com/DrewThomasson/ebook2audiobook/issues/631 though

If that helps

ROBERT-MCDOWELL:
fixed on v25.4.25. thanks!

PR #656, opened 2025-03-31
Fix finetune kaggle + colab link

DrewThomasson:


DrewThomasson:
@ROBERT-MCDOWELL 👍 

sorry about this mess but it's only touching the readme and Notebooks folder anyway

PR #658, opened 2025-04-01
Add Manual Python install guide

DrewThomasson:


DrewThomasson:
Was contemplating on this but unsure If it's a good idea at all....

DrewThomasson:
On second thought, This seems like it'll cause more issues ...

PR #766, opened 2025-05-30
Docker compose patch 1

DrewThomasson:
Waiting for verification rom @yodatak


DrewThomasson:
yup works in testing 👍 
@ROBERT-MCDOWELL Merging into v25 then now

PR #768, opened 2025-05-31
fixing docker compose and adding podman compose

DrewThomasson:


DrewThomasson:
@ROBERT-MCDOWELL 
Already confirmed to work on my end and in my testing so merging changes into main 👍

PR #771, opened 2025-06-01
Adding more dynamic cuda pytorch selector

DrewThomasson:


DrewThomasson:
This would allow someone to build for any cuda version easily in docker

It'll dynamically search for a nightly or a stable build of what ever cuda number they put in

Running Dev test on this first...

PR #776, opened 2025-06-02
Revert "Add inline comment to app.py"

ROBERT-MCDOWELL:
Reverts DrewThomasson/ebook2audiobook#769


ROBERT-MCDOWELL:
thanks but no need to comment this function it's enough explicit.

PR #833, opened 2025-06-14
Update lang.py

DrewThomasson:


DrewThomasson:
Added a few english abbreviations I could remember off the top of my head :)

PR #844, opened 2025-06-17
Create translate-readme.yml

DrewThomasson:


DrewThomasson:
Dang it its not translating them its just returning the untranslated readme files but re-named.....

PR #924, opened 2025-07-14
Deprecated readme removal

DrewThomasson:
Removing the other readme translations for now as they are causing issues with deprecated commands and guides. + causing confusion for no reason

Once I figure out how to update them I'll be able to put something in its place


DrewThomasson:
kk merging then

PR #930, opened 2025-07-16
quick fix docker-build+test.yml

DrewThomasson:
This is quick fix to make it always give the latest docker images the correct tag.

As sometimes it... does not

This is because it is allowed to push and build the varients even... if the base builds never build... which pushes the.. old cached version... with the new version tag...

Now it requires the base build to push as the latest and as the tag suchlike.

Which should fix this issue while I rebuild and streamline the workflows further..

### DIFF
--- a/.github/workflows/docker-build+test.yml
+++ b/.github/workflows/docker-build+test.yml
@@ -270,7 +270,7 @@ jobs:
 
 
   final_push:
-    needs: run_tests
+    needs: build
     if: ${{ !failure() || inputs.require_all_tests == false }}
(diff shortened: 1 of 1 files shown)


DrewThomasson:
Note to Drew:

You may have to also link up multiple workflows

like make one that will only run tests

and call on that in other workflows to re-use for instance the test workflow in multiple so i dont have to repeat things

I could also reuse the build base workflow and such too like this.

hopefylly this fixes the paralell test issues I am running into,,, cause its marked as one test but for some reason it doesnt work in paralell idk

PR #962, opened 2025-07-20
Fix two voices fine-tune + add extra voice samples

DrewThomasson:
added the the missing slash


DrewThomasson:
Had to remove the SubZero fine-tune, 

Not good enough, became overFit after training on 8 epoches

DrewThomasson:
We can remove any that have too much noise or you think needs modifications done to the audio

ROBERT-MCDOWELL:
more the audio is clear,  with presence and no noise, better it is indeed.
you can use the normalize script.

DrewThomasson:
Does E2A not already normalize it?

ROBERT-MCDOWELL:
> Does E2A not already normalize it?

are you talking about the fine tuning or the 16khz 24khz voice you create in ./voices ?

PR #970, opened 2025-07-22
Add set -e at beginning of ebook2audiobook.sh

DrewThomasson:
@ROBERT-MCDOWELL 

This will allow `ebook2audiobook.sh` to give the error code the app.py will give if it fails

fixing My local workflow issue :)


ROBERT-MCDOWELL:
if it can help you so not prob!

PR #979, opened 2025-07-23
Super Duper Workflow Optomization stuff very good

DrewThomasson:


DrewThomasson:
The Previous version was VERY bad

So much for Vibe Coding

PR #994, opened 2025-07-23
V25

DrewThomasson:


DrewThomasson:
Fixed the two Workflows, They now simulate Tests on the correct branch or SHA

PR #995, opened 2025-07-23
Break test

DrewThomasson:
This should break in this PR as a test to see if its checking out correctly 

and the requirments.txt should be gone locally on the runner while this is running


ROBERT-MCDOWELL:
requirements.txt must not be deleted at all indeed

DrewThomasson:
Oh don’t worry that was a note for myself for what I should see locally when this PR is tested

DrewThomasson:
This was just to test to see if the auto Test on any PR to the main from any branch would be tested correctly

PR #1003, opened 2025-07-25
Filter out unusually short chapter blocks using statistical outlier detection.

DrewThomasson:
> This commit adds statistical outlier filtering to the chapter splitting logic. 

After chapters are detected, any chapter block that is unusually short (based on a configurable standard deviation threshold) is automatically removed. 

This is my attempt at removing the junk blocks that are too short

- The filter uses the mean and standard deviation of chapter lengths to determine outliers.
- The cutoff is adjustable with the `stddev_factor` parameter.

### DIFF
--- a/lib/functions.py
+++ b/lib/functions.py
@@ -18,6 +18,7 @@ import platform
 import random
 import shutil
 import socket
+import statistics
 import subprocess
 import sys
 import tempfile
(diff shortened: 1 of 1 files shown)


DrewThomasson:
Testing PR in this workflow here

https://github.com/DrewThomasson/ebook2audiobook/actions/runs/16525602975

Edit: Passed

ROBERT-MCDOWELL:
I planned to develop a pre validation of detected block the user will select or not before to start the conversion.
unfortunately it can be a "chapter" with only few words, chapter being a title of a very small poem for example.

DrewThomasson:
Could we use this type of logic in another way perhaps?

Say, auto detecting and highlighting what "chapters" are most likely to be not real, 

or even setting them by default to be "un-selected" 🤔

PR #1036, opened 2025-08-08
Add Kokoro

DrewThomasson:


DrewThomasson:
Turns out Kokoro has different voices per Language,

I am having trouble figuring out how to make the voice cloning dropdown populate with the appropriate kokoro voices per selected kokoro supported language

PR #1063, opened 2025-09-10
Correcting kab

BoFFire:
Hi,

Correcting the name an native name for `kab` language.


ROBERT-MCDOWELL:
thanks I'm correcting it manually directly into the lang.py as it needs to respect the sort order.

Issue #1067, opened 2025-09-19
The app only generates the audio of a little part of the epub.


nandosssssss:
This is just a rudimentary guide, don't be too intimidated, You don't have to follow this to the tee
BEFORE TO OPEN AN ISSUE, MAKE SURE THAT IT HASN'T ALREADY BEEN SOLVED OR REPORTED. :)
Check it with the issue search field

Script Mode

Gradio/GUI

Script Version

25.8.18

Process Mode

Operating System:

Windows 11

Command Line used to start 

ebook2audiobook.cmd

Details
Describe the bug

The app only generates the audio of a little part of the epub.

It sounds well and uses well the reference audio, but only with a small part of the epub.

I have tried with several books, in different formats: epub mobi txt, etc

I have closed the terminal and the browser many times. And I have restarted the computer.

You can find the terminal log here:
https://privatebin.net/?6c01e68912873edb#6MB4gtYhoK9Kdr63Vo7QzdtKQULiPU5VZXM3vq9AQunm

Thanks

nandosssssss:
Thanks for the help. I’ll close the post

ROBERT-MCDOWELL:
your issue is indeed a duplicate of #1066 which I fixed for the next release.

nandosssssss:
I think it’s no the same error. That post talks about “ I noticed a problem when the text contains times in such a format as, for example, 00:00 , 12:30 ,. . .” but I have problems with every epub I tried. It begins to generate the audio using a reference but it always stops after several phrases. The audio generated is almost perfect but it only generates some phrases

nandosssssss:
I think it’s no the same error. That post talks about “ I noticed a problem when the text contains times in such a format as, for example, 00:00 , 12:30 ,. . .” but I have problems with every epub I tried. It begins to generate the audio using a reference audio but it always stops after several phrases. The audio generated is almost perfect but it only generates some phrases 

May be I am doing something wrong. I ave tried with different EPUBs even converting them into other formats like txt and so. 

I don’t want to bother you. Your app is great anyway. I just wanted to hear my epubs with the voice I like most. Many thanks

ROBERT-MCDOWELL:
yes it's the same, the logs show it clearly. having this issue on time formatted lik xx:xx so some sentences following it can miss...

nandosssssss:
Ok. Many thanks

ROBERT-MCDOWELL:
[...]

PR #1098, opened 2025-10-23
Fix/docker build failure

Alan-Cheng:
### Problem
The Dockerfile failed to build because the versions of pandas and numpy caused compilation errors during installation.

### Solution
- Pin numpy to a compatible version in requirements.txt
- Pin pandas to a version compatible with the pinned numpy
- Removed duplicate package specifications in Dockerfile

### Testing
- Built the Docker image locally using the fixed Dockerfile
- Verified that `docker build .` completes without errors

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,8 +24,7 @@ ENV PATH="/root/.cargo/bin:${PATH}"
 # Copy the application
 WORKDIR /app
 COPY . /app
-# Install numpy and pandas first to avoid compilation issues
-RUN pip install --no-cache-dir "numpy<2.0.0" "pandas==2.2.3"
+# Note: Version constraints are managed in requirements.txt
 
 # Install UniDic (non-torch dependent)
 RUN pip install --no-cache-dir unidic-lite unidic && \


Alan-Cheng:
@DrewThomasson Could you please review this PR? It fixes the Dockerfile build issue. Thank you!

DrewThomasson:
I'll check later today but I don't think the requirments.txt file needs to be modified

DrewThomasson:
@Alan-Cheng Test to see if your build actually functions, in generating a small audiobook 

Iv'e had issues with that before

Alan-Cheng:
Thanks for the suggestion! I've built an updated version and tested it by converting a ~2.2 MB book. The generation is currently about 30% complete, and the dependency issue that previously caused the image build to fail hasn’t reappeared.

This version includes the changes from PR #1099 , which fixes issue #1067  and allows the job to continue generating properly.

DrewThomasson:
I'll look into merging this and such once we get a the main E2A program working again

DrewThomasson:
These modifications may actually work better

As I'm also streamlining the Dockerfile for other issues and such

(Keep in mind v25 branch is broke at the moment so if you want to try this you would need to apply these modifications to the main)

https://github.com/DrewThomasson/ebook2audiobook/pull/1105

DrewThomasson:
Probs not going to merge as our upcoming update will fix this issue along with others like getting correct torch version and such like

I'll keep this open until the next release tho

DrewThomasson:
Closing as next release is coming out

DrewThomasson:
the new release is out, closing this then

PR #1107, opened 2025-10-29
Revise Readme Docker section and improve clarity

DrewThomasson:
Updated Docker section in README to clarify options and recommendations.


DrewThomasson:
This is me attempting to put Docker Compose as Priority

Not sure if this makes it easier or harder for newcomers

But The compose does run into less issues typically, and is easier for us to debug for them..... hm

DrewThomasson:
I've now only the docker compose instructions in the readme

If someone wants to use it manually with docker run then they can go to the new wiki for that

PR #1113, opened 2025-11-05
first Working E2A launcher on mac

DrewThomasson:
@ROBERT-MCDOWELL 
This successfully creates a launcher on my mac

Unknown how to go about creating a linux script or other distros but this looks like a good foundation for us to bounce off of 

I included the other things like the original base `appLogo.png`, the `generate-icons.py` icons script used, and the generated icons for `linux`, `mac`, and `windows` 

What's your thoughts?  :)

The mac launcher just opens the users terminal, runs conda deactivate command (just in case), runs the `ebook2audiobook.sh`, waits 10 seconds and then opens `http://localhost:7860/` on the default browser


ROBERT-MCDOWELL:
why not!

PR #1115, opened 2025-11-05
Add PeterGriffanFamilyGuy xtts fine-tune

DrewThomasson:


DrewThomasson:
Adding A Fine-tuned Xtts model for this

PR #1118, opened 2025-11-08
move docker section into usage

ndzn:


DrewThomasson:
Oh thx that looks better

PR #1124, opened 2025-11-11
Enhance E2A-Test workflow with custom model test

DrewThomasson:
Updated the E2A-Test workflow to include a new custom fine-tuned XTTSv2 model test and added a new output folder for XTTSv2FineTuneCustom.


DrewThomasson:
looks good Merging

PR #1127, opened 2025-11-11
Implement early return for headless mode and existing bundle

DrewThomasson:
Added checks for headless mode and existing app bundle.


DrewThomasson:
@ROBERT-MCDOWELL 
kk appears fixed on my end

Merging into `v25`

PR #1139, opened 2025-11-17
Log and remove __sessions directory in workflow

DrewThomasson:
List the contents of the __sessions directory before removing it.


DrewThomasson:
@ROBERT-MCDOWELL 
Here you go

Now itll do `ls -lhR ./models/__sessions`

before it does the `rm -rf ./models/__sessions`

so then we can see everything that is in the ./models/__sessions folder before it's wiped

DrewThomasson:
@ROBERT-MCDOWELL 
Merging

PR #1197, opened 2025-11-24
Update README with Podman Windows startup issue

DrewThomasson:
Added a note about Podman startup issues on Windows.

### DIFF
--- a/podman-compose.yml
+++ b/podman-compose.yml
@@ -28,6 +28,7 @@ services:
       - ./:/app  # Maps the local directory to the container.
 
 # Common Issues: ----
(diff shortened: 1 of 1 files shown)


DrewThomasson:
hm, Should I just put it in the podman compose file itself and not on the Readme?

ROBERT-MCDOWELL:
podman compose is almost identical to standard compose so since Im' working on complete different and clean context with no redundant code hard to maintain with the installer sh script, Dockerfile, compose etc... so podman could and should always be a copy of standard compose with the incompoatibilities removed and changed with the podman settings.

DrewThomasson:
I suppose I should close this given your docker and compose modifications?

ROBERT-MCDOWELL:
yes, I will check again if the standard compose is podman compatible so no need to have another file.

PR #1226, opened 2025-11-28
Add session recovery mechanism for handling computer sleep

TheCamer1:
Hibernating or sleeping your computer leads to the following error:

```
expected str, bytes or os.PathLike object, not NoneType
  File "/app/lib/functions.py", line 1789, in combine_audio_sentences
    chapter_audio_file = os.path.join(session['chapters_dir'], chapter_audio_file)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen posixpath>", line 76, in join
TypeError: expected str, bytes or os.PathLike object, not NoneType
Caught DependencyError: expected str, bytes or os.PathLike object, not NoneType
combine_audio_sentences() failed!
```

This means you have to start the whole audio generation again. This PR allows the code to recover by inspecting the file system to populate the state again.

### DIFF
--- a/lib/functions.py
+++ b/lib/functions.py
@@ -1615,6 +1615,20 @@ def convert_chapters2audio(id:str)->bool:
             msg = 'Cancel requested'
             print(msg)
             return False
+            
+        # Attempt to restore session directories if they were lost (e.g., after hibernation)
+        if session['chapters_dir'] is None or session['chapters_dir_sentences'] is None:
+            if not restore_session_directories(session):
+                error_msg = "Session state lost and could not be restored. This can happen after hibernation or system interruption. Please restart the audiobook generation process."
+                print(error_msg)
+                return False
+        
+        # Verify critical session data still exists after restoration
+        if session['chapters'] is None:
+            error_msg = "Critical session data (chapters) was lost. The parsed text content is missing. This indicates severe session corruption. Please restart the audiobook generation process."
+            print(error_msg)
+            return False
+            
         tts_manager = TTSManager(session)
         resume_chapter = 0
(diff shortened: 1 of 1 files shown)


ROBERT-MCDOWELL:
for each PR thanks to provide:
- the version you patched
the error you see is simple, the hibernation, sleep went beyond the limit of the gradio cache which wipe out after 24h.
so this is a normal behavior.
about resume  a conversion, this is something solved since almost one year, and to profit from it you just need to reload the page with the same browser and session and import the exact same file with at least the same TTS engine.
And again, please don't send PR made by A.I. code. you should understand what you are coding.

ROBERT-MCDOWELL:
I reproduced the python issue above, it has nothing to do with hibernation or sleep but browser stalls or interruption.
this is something happening randomly and it's not easy to debug and solve it.

ROBERT-MCDOWELL:
Correction: it comes from the session which is a dictProxy dict() working between threading and multiprocessing tasks having some trouble to update the value of session rows at the right time (as it's async between threads) when the CPU/GPU is very busy or the OS itself is slowing down, sending to the function (combine_audio_sentences() in this case) a None value.
This case is flagrant when you modify the argument itself in its own function in a for/while loop.
renaming the var in the function where it's concatenated with session['chapters_dir'] solve the issue as session['chapters_dir'] never changes as long as the session exists in the browser (even after quit and reload) and the backend.
in fact hibernation or sleep but also busy CPU/GPU/OS are the triggers for this bug.

PR #1227, opened 2025-11-28
Improve TTS text preprocessing to prevent audio artifacts

TheCamer1:
This PR fixes four things:

- The xtts engine has an issue where it will occasionally repeat the end of a long sentence. During testing, I believe the cause of this is because full stops are replaced with dashes in a sentence to indicate to the model to pause. But the code was replacing full stops including the final full stop of a sentence with a dash, and I believe this caused the model to hallucinate that the sentence needed to continue, and so it repeated the last few words. After making this change, I haven't had it hallucinate like this at all.
- The xtts engine was occasionally hallucinating around quotation marks, merging words together across the quotation boundary and mispronouncing words. To fix this I have removed the first quotation when it appears and replaced the second with a comma, which indicates that the model should pause. I've found this gives it much better performance when saying a quotation
- Sentence splitting had a bug where if a sentence ended with a quotation like "this is the end of the sentence.", it would not actually recognise that that full stop was the end of the sentence at all, and just continue on finding a break somewhere else, leading to some very unnatural sentence flow. I've had to preserve quotes when calculating sentence boundaries to fix this (reason for lang.py changing). 
- Sentence splitting had another bug where it would occasionally produce very short sentences of even just one word when a sentence got too long, just splitting a sentence up unnaturally when it hit the sentence character limit. I've added code to discourage it from doing this, and a post processing step to guarantee this won't happen if any sneak through.

### DIFF
--- a/lib/classes/tts_engines/coqui.py
+++ b/lib/classes/tts_engines/coqui.py
@@ -450,6 +450,29 @@ class Coqui:
         sf.write(tmp_path,wav_numpy,expected_sr,subtype="PCM_16")
         return tmp_path
 
+    @staticmethod
+    def _clean_text_for_tts(text:str)->str:
+        """
+        Cleans text for TTS by handling quotes and punctuation to prevent pronunciation issues.
+        - Preserves final period to prevent hallucination/repetition
+        - Removes opening quotes, replaces closing quotes with commas for natural pauses
+        """
+        text = text.rstrip()
+        
+        # Keep final period, replace internal periods with em dash for natural pauses
+        if text.endswith('.'):
+            text = text[:-1].replace('.', ' —') + '.'
+        else:
+            text = text.replace('.', ' —')
+        
+        # Handle quotes: opening removed, closing becomes comma (only if not already punctuated)
+        text = re.sub(r'(?<![,.;:!?])("(?=[\s,.;:!?\)—]|$))', ',', text)
+        text = text.replace('"', '')
+        text = re.sub(r"(?<![,.;:!?])('(?=[\s,.;:!?\)—]|$))", ',', text)
+        text = text.replace("'", '')
+        
+        return text
+
     def convert(self, sentence_index:int, sentence:str)->bool:
         try:
             speaker = None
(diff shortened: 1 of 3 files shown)


ROBERT-MCDOWELL:
is your PR tested with all XTTSv2 languages? if some punctuations are replaced by other it's actually because of to avoid hallucination and unwanted result. And it's crucial to precise the language as each language needs a different treatment. I don't talk about other engines like VITS, FAIRSEQ etc... which ask also another condition.
Most of what yiou said is fixed since long time. which version  are you using?
Also create a PR from A.I. code is just a no, no. you should understand the code your are generating and cleanup its comments and have a final code with strict typing.
split sentences and punctuations are something I'm working on since almost one year as each language, engine and model has its own particularity,

ROBERT-MCDOWELL:
btw all the code the A.I. did in this PR is gross, dirty and create a mess into the class and functions.py, and most of the issues exposed are fixed since very long time.

TheCamer1:
Apologies if the code isn't to your liking, I would have been willing to make it to your liking. I do understand what I'm coding, I have extensive experience with Python, I simply used AI to help. The four issues I had above are on the latest tip of develop, v25.11.25. I could repeat these bugs consistently, I wouldn't be so quick to dismiss them. At the very least my code can give you an idea of the solution, because the code fixes the bugs, even if you want to write the code to fix them from scratch. In particular the sentence splitting bugs were quite problematic, leading to consistent mistakes where sentences would be split up in the middle, and you'd get one word sentences in the output taken from the middle of a real sentence.

ROBERT-MCDOWELL:
the issue is not if I like or not the coding of this PR. it's more than E2A has it's own normalization and must be respected.
for that, the PR owner should know how E2A is normalized by just reading the code. for example single quotes are used in all code but not in dict. strict typing, etc....
there is a big difference between AI help and copy/paste, without to know how the code is organized  and contextualized.
and again, you should spcifiy:
- which language are causing this.
- which tts engine are causing this
- which settings (very important with Bark and Xttsv2)
- the model you are using (your own model or fine tuned presets can behave very differently than the builtin model)

E2A manages +1110 languages and more than 7 TTS engines and more coming. so you can imagine apply a patch without being tested on all cases cannot be considered as valuable.
and btw the code generated by this AI is not optimized at all and does not take the context in consideration (A.I need the entire project to target the right context). I lost enough time to trust A.I. for coding. Indeed I use it for some memory recall or very small piece of code, but always after exposing the context. or it's a wasting time nightmare.

ROBERT-MCDOWELL:
it's better you open an issue, give ALL informations needed (or I will close the issue), provide the language, the E2A version (btw v25.11.25 is not on the release list anymore for some reasons so you should downgrade to v25.11.22) the tts engine used and a full terminal logs of the conversion, providing the text so I can reproduce it and confirm there is an issue.

PR #1350, opened 2025-12-16
Remove deepspeed installation under Windows

chshkhr:
DeepSpeed is not needed (and not practical) on Windows because:

❌ No NCCL support

❌ No GPU distributed training

❌ Custom CUDA ops fail to build

❌ Windows toolchain incompatibility

✅ PyTorch already covers common Windows use cases

Use DeepSpeed only on Linux (or WSL2).


ROBERT-MCDOWELL:
I'm in contact with deepspeed dev to have more info about it. once I will have the answer I will comment it or modify the code if there is a way to compile.

ROBERT-MCDOWELL:
we removed completely deepspeed as it's deprecated from their authors with the recent version of torch.

PR #1410, opened 2025-12-25
Simplify Docker-Build.yml

DrewThomasson:
Makes the huggingface build only run if the first 2 docker builds succeed


DrewThomasson:
@ROBERT-MCDOWELL 
That should make the Docker Build Not run if the E2A Test Fails

Merge 👍 ?

ROBERT-MCDOWELL:
Yes!

PR #1435, opened 2025-12-27
Create root .gitignore

mrienstra:
Guessing all of these files should be ignored?


DrewThomasson:
Looks good to me what do you think @ROBERT-MCDOWELL?

ROBERT-MCDOWELL:
sure! I thought it was already done!

PR #1442, opened 2025-12-28
Add Hungarian VITS Custom-Voice test to workflow

DrewThomasson:
@ROBERT-MCDOWELL 
How does this look?


ROBERT-MCDOWELL:
It passed but I suspect the voice conversion to the language missing in headless mode. :**(

DrewThomasson:
?

ROBERT-MCDOWELL:
it seems the eng -> xxx voice conversion to use the voice cloning in the same language is running only in gradio mode. I must check it.

DrewThomasson:
ah kk 👍

ROBERT-MCDOWELL:
you can merge as the possible issue is out of the workflow anyhow

PR #1512, opened 2026-01-11
fixes google colab

BFG-BFG:
fixes google colab but less pretty and might break if google changes cuda version, not sure how efficient it is


DrewThomasson:
You can confirm the GPU works and not only the CPU?

ROBERT-MCDOWELL:
@BFG-BFG we are waiting your answer thanks

BFG-BFG:
nope it's all broken again

PR #1656, opened 2026-03-07
fix(gradio): skip invalid restored slider values

ElhamDevelopmentStudio:
## Summary
Fixes a Gradio crash when processing conversion after slider values are restored from browser session storage.

## Root cause
Slider restore JS parsed missing/invalid saved values and wrote them into numeric inputs.
That could produce `None` payloads in Gradio preprocess, causing:
`TypeError: '<' not supported between instances of 'NoneType' and 'int/float'`.

## Changes
- Updated slider restore logic in `lib/gradio.py`:
  - skip restore when saved value is missing/empty
  - parse safely (`parseInt(..., 10)` / `parseFloat(...)`)
  - apply only finite values (`Number.isFinite(...)`)
- Leaves default slider values untouched when saved values are invalid.

## Validation
- `python3 -m py_compile lib/gradio.py`
- Manual: restart app, open UI, switch tabs, run conversion without slider preprocess errors.

### DIFF
--- a/lib/gradio.py
+++ b/lib/gradio.py
@@ -2402,9 +2402,18 @@ def build_interface(args:dict)->gr.Blocks:
                                         ];
                                         sliders.forEach(slider => {
                                             if(!slider) return;
-                                            const key = slider.closest("div[id]").id.replace(/^gr_/, "");
-                                            const saved = window.session_storage[key];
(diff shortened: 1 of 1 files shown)


ROBERT-MCDOWELL:
Thanks!

PR #1710, opened 2026-04-12
Jt/english year handling

jamestufen:
Making english language year pronounciation a bit better.

I kept having years like 1845 transcribed as "One thousand, eight hundred, and forty-five" instead of "eighteen-forty-five". This should catch most of those. English only, so I tried to keep the logic in its own file.


ROBERT-MCDOWELL:
this cannot be accepted, the code is too messy and no more file should be created just for that.
btw there is already a language classification from conf_lang.py "year_to_decades_languages"
which is managed by the the stanza model taking care of this date behavior.
the issue here is maybe stanza model couldn't be downloaded so num2words which is not perfect for english took the relay?
better you open an issue and provide the text.

jamestufen:
Totally understand, this is a lot of code for one specific issue. 

The root issue is that in core.py, it checks for a year (e.g. 1999) and an ordinal (e.g. 23rd) before kicking off the date replacement. Lots of years get missed by this, e.g. the sentence "From 1949 to 1976, Mao ruled the country" would never hit the date logic [here](https://github.com/DrewThomasson/ebook2audiobook/blob/7c47941aefed84c4ecc38d0c25910ca77bf732f7/lib/core.py#L1135), and instead be pronounced "From one thousand, nine hundred, and forty-nine..."

Maybe the solution is just to have stanza run on all chunks looking for date, but that seems inferencially costly. I'll write up an issue and link it.

jamestufen:
https://github.com/DrewThomasson/ebook2audiobook/issues/1711

ROBERT-MCDOWELL:
stanza is not costly and run since one year very well. it takes care of everything you are talking about. as I said you run the app and something happened stanza model didn't download or else.

ROBERT-MCDOWELL:
not only to mention that stanza model is taking care of more than 80 languages...

ROBERT-MCDOWELL:
"From 1949 to 1976," is a good example of why a model treating the date carefully is mandatory. how you will differentiate 1949 as a year and "1949 breads sold" or "1949 x 78" ?

jamestufen:
@ROBERT-MCDOWELL , did some more testing, and you were right, my old fix was way too complex. This is a much cleaner fix- we're just running stanza against a wider selection of chunks. Running on only chunks with ordinals misses a lot.

To see for yourself, compare the output of main and my branch for this simple sentence:
```
echo 'He was born on August 22, 1904 and died on February 19, 1997.' > /tmp/year_test.txt
git checkout main
./ebook2audiobook.command --headless --ebook /tmp/year_test.txt --output_format m4b --language eng --output_dir /tmp/main_output
# he was born August two twenty one thousand nine hundred and four, and died on February, one hundred and ninety-one thousand, nine hundred ninety-seven
git checkout jt/englishYearHandling
./ebook2audiobook.command --headless --ebook /tmp/year_test.txt --output_format m4b --language eng --output_dir /tmp/fix_output
# He was born on August twenty-second, nineteen oh four, and died on February nineteen, nineteen ninety-seven
```

Let me know what you think appreciate your feedback!

ROBERT-MCDOWELL:
looks good indeed. just provide the patch here and will integrate it in the coming v26.4.12

jamestufen:
Here's the patch — one line change in `lib/core.py` at line 1086:

```diff
-                if re_num.search(text) and re_ordinal.search(text):
+                re_year = re.compile(r'\b(?:1[0-9]|20)\d{2}\b')
+                if re_num.search(text) and (re_ordinal.search(text) or re_year.search(text)):
```

Thanks!

ROBERT-MCDOWELL:
patched! thanks!

PR #1757, opened 2026-04-24
Add a [chapter] sml tag (Concept)

DrewThomasson:
Seem to work on my end, 

Add manual [CHAPTER] tag splitting and fix audio stitching crash

But in doing so I had to add a fix for a bug where the chapter might start with "." so its skipped but that ... crashed it?

AI Explanation

```
Add manual [CHAPTER] tag splitting and fix audio stitching crash

Tested and appears to work well on my end! Added the ability to manually cut chapters using a [CHAPTER] tag in the text.

But in doing so, I had to fix a pre-existing bug where a chapter block might start or end with just a . (or other stranded punctuation).

Because a solitary . has no letters or numbers, the audio generator correctly skipped rendering a .flac file for it. However, the 

combine_audio_sentences()
 function blindly expected a file to exist for every single line and crashed when it couldn't find the audio file. I synced up the logic so the audio stitcher now properly ignores blank/punctuation-only lines just like the generator does.


```

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,7 @@ https://github.com/user-attachments/assets/81c4baad-117e-4db5-ac86-efc2b7fea921
 - `[break]` — silence (random range **0.3–0.6 sec.**)
 - `[pause]` — silence (random range **1.0–1.6 sec.**)
 - `[pause:N]` — fixed pause (**N sec.**)
+- `[chapter]` — chapter marker
 - `[voice:/path/to/voice/file]...[/voice]` — switch voice from default or selected voice from GUI/CLI
 
 **Check our other repo dedicated to add SML automatically in your ebook -> [E2A-SML](https://github.com/DrewThomasson/E2A-SML)**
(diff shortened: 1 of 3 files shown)


DrewThomasson:
I don't expect this to actually be merged but I thought it was a interesting concept wise

ROBERT-MCDOWELL:
do you mean in case of a raw text without structure s [chapter] will be used?

ROBERT-MCDOWELL:
so [chapter:{title}] xxxxxx [/chapter]

DrewThomasson:
> do you mean in case of a raw text without structure s [chapter] will be used?

Yes yes yes exactly

DrewThomasson:
> so [chapter:{title}] xxxxxx [/chapter]

That a better format

I forgot about the title and chapter number part

ROBERT-MCDOWELL:
well, chapter number could be the order of the chapter mark. so no need for chapter number.

ROBERT-MCDOWELL:
it can be used also like
[chapter]the first sentence is always the title. hellow world! this is the first chapter blablabla[/chapter]
but more robuse to use [chapter:this is a great title] xxxx [/chapter]
do you need it now?

DrewThomasson:
yeah, It's for the https://github.com/DrewThomasson/E2A-SML

but there's no rush as it's not required

DrewThomasson:
Closing this PR then

PR #1783, opened 2026-04-27
Make tortious only test with a single file instead of batch

DrewThomasson:


DrewThomasson:
@ROBERT-MCDOWELL 
This should cut the full E2A test time in HALF :).    

Merge? 👍?

DrewThomasson:
Merging

PR #1790, opened 2026-04-29
fix: build prev_blocks before voice conversion to prevent false block changed on resume

### LINKED ISSUE #1789: Resume broken for non-English languages: block_changed=True on every restart
## Bug

When resuming an interrupted conversion with a non-English language,
all blocks show `changed, reconverting` and the conversion restarts
from block 0, losing all progress.

## Root Cause

In `convert_chapters2audio()`, the voice conversion loop (`set_voice()`)
runs **before** `prev_blocks` is built. For non-English languages,
`set_voice()` transforms the block voice path to an internal session path:

    voices\deu\voice.wav → voices\__sessions\voice-<uuid>\deu\voice.wav

Since `block_hash()` includes the `voice` field, and `blocks_current`
has the transformed path while `blocks_saved` has the original,
hashes never match → `block_changed=True` for every block.

## Fix

Move `prev_blocks` construction to before the voice conversion loop.

## Reproduction

1. Start conversion with `--language deu` and a custom `--voice`
2. Interrupt mid-way
3. Resume with `--session <id>`
4. All blocks show `changed, reconverting`

Tested: v26.4.27, Windows 11, CUDA, `--language deu`

thekk1:
Fixes #1789

For non-English languages, set_voice() transforms the block voice path to an internal session path before the hash comparison. This caused block_changed=True on every resume even when nothing had actually changed, forcing all blocks to be reconverted from scratch.

Fix: move prev_blocks construction to before the voice conversion loop, so both current_hash and hash_ref are computed with the original voice path.

### DIFF
--- a/lib/core.py
+++ b/lib/core.py
@@ -2182,6 +2182,7 @@ def convert_chapters2audio(session_id:str)->bool:
         sentence_resume = blocks_current['sentence_resume']
         if session['cancellation_requested']:
             return False
+        prev_blocks = {b['id']: b for b in (session.get('blocks_saved') or {}).get('blocks', [])}
         xtts_languages = default_engine_settings[TTS_ENGINES['XTTSv2']].get('languages', {})
         if session['language'] != 'eng' and session['language'] in xtts_languages:
             voice_cache = {}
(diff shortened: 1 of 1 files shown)


ROBERT-MCDOWELL:
correct! fixed for the next update, thanks!

ROBERT-MCDOWELL:
you can submit PRs to my dev repo it will be easier for me to manage.
https://github.com/ROBERT-MCDOWELL/ebook2audiobook

PR #1792, opened 2026-04-29
fix: append [break] token to preceding text element to prevent sentence-start distortion

### LINKED ISSUE #1791: [break] token at sentence start causes TTS distortion on first word
## Bug Description

When converting ebooks, `[break]` tokens appear at the **beginning**
of sentences in the TTS input instead of at the end of the preceding
sentence. This intermittently causes audible distortion, crackling or
hallucination on the first word of the affected sentence.

The distortion is most noticeable on **short sentences that form their
own paragraph**, as the TTS model (XTTSv2) has little context to
stabilize the inference when the first token is `[break]` followed by
only a few words. Longer sentences are less affected because the
subsequent text provides enough context to recover.

## Example

Conversion log shows:
: [break]Dann begann sie, ihre Sachen zu packen.
: [break]Brauchst du Hilfe?

Short single-sentence paragraphs like `Brauchst du Hilfe?` are
particularly prone to distortion on the first word.

## Root Cause

In `filter_blocks()`, `[break]` is inserted as a **separate element**
between text chunks in `text_list`. After `' '.join(clean_list)`:

    "...warf. [break] Dann begann sie..."

When `get_sentences()` splits at `.`, `[break]` lands at the start
of the next sentence:

    sentence 1: "...warf."
    sentence 2: "[break]Dann begann sie..."

## Expected Behavior

`[break]` should be appended to the **preceding** sentence so it is
processed at the end of an inference call rather than the beginning
of the next one:

    sentence 1: "...warf.[break]"
    sentence 2: "Dann begann sie..."

## Fix

Append `[break]` directly to the last text element in `text_list`
instead of inserting it as a separate entry. If `text_list` is empty
or the last element is itself a SML token, fall back to the current
behavior.

## Reproduction

1. Convert any ebook with any language
2. Observe sentences starting with `[break]` in the conversion log
3. Listen to short single-sentence paragraphs — first word is
   intermittently distorted

Tested: v26.4.27, Windows 11, CUDA, language=deu

thekk1:
Fixes #1791

[break] tokens were inserted as separate elements between text chunks during HTML parsing. After sentence splitting, they ended up at the beginning of the following sentence. XTTSv2 then processed [break] as the first token of a new inference call with no prior speech context, causing distortion or hallucination on the first word.

Fix: append the [break] token directly to the last text element in text_list (if it is not itself a SML token), so the break is processed at the end of the preceding inference call instead.

Before: "[break]Dann begann sie..."
After:  "Dann begann sie..."  (break appended to previous sentence)

### DIFF
--- a/lib/core.py
+++ b/lib/core.py
@@ -1202,7 +1202,11 @@ def filter_blocks(session_id:str, idx:int, doc:EpubHtml, stanza_nlp:Pipeline, is
                     text_list.append(payload.strip())
                 elif typ in ('break', 'pause'):
                     if prev_typ != typ:
-                        text_list.append(sml_token(typ))
+                        token = sml_token(typ)
+                        if text_list and text_list[-1] not in {v['static'] for v in TTS_SML.values() if 'static' in v}:
+                            text_list[-1] = text_list[-1] + token
+                        else:
+                            text_list.append(token)
                 elif typ == 'table':
                     table = payload
                     if table in handled_tables:


ROBERT-MCDOWELL:
good catch! patched on the dev repo so it will be available on the next update. thanks!

thekk1:
> good catch! patched on the dev repo so it will be available on the next update. thanks!

Does that mean I should target my PRs against a different branch?

ROBERT-MCDOWELL:
no it's ok, I just did it manually on my side. we have a little complex dev repo chain but it's ok. you can continue to PR here so  I will manually patch it.

PR #1805, opened 2026-05-01
Local patches for CachyOS Linux + CUDA install

Retro-Jack:
- ebook2audiobook.command:
  * TMPDIR override to /tmp/e2a (AF_UNIX 108-char path-length fix for long install paths)
  * LD_LIBRARY_PATH prepend for site-packages/nvidia/*/lib so torchcodec's bare ctypes.CDLL() finds the CUDA libs (e.g. libnppicc.so.13)
  * Auto-install nvidia-npp wheel on CUDA installs (torchcodec dep, not pulled by upstream torch install set)
  * Interactive Tesseract OCR language-pack picker on Arch (after main pacman install loop, skips on non-TTY)
- lib/conf.py: TMPDIR/tempfile.tempdir override at Python layer (the bash export alone wasn't enough; conf.py was overwriting it back to run/)
- lib/core.py: prepend /usr/bin to PATH for ebook-convert subprocess so shebang resolves to system Python (with python-mechanize) instead of conda env Python; also log returncode + stderr on failure
- start.sh: convenience launcher that polls :7860 and opens the browser
- Docs/: install/usage/patches/troubleshooting/moving documentation
- VOICES.md / VOICES.odt: descriptions of every shipped voice sample
- .gitignore: ignore .installed, /nul, " Moving.txt"


ROBERT-MCDOWELL:
oh my..... I can't stand anymore all these A.I. stupid PR....

ROBERT-MCDOWELL:
ok let's say I apply your patch. what about Windows, all other linux distrib, macOS X ?

PR #1827, opened 2026-05-05
Fix crash when mwt processor is unavailable in Stanza

AddeSz:
Some languages (for example: Swedish) cause the pipeline to fail because the mwt processor is requested even though it is not available in Stanza's resources for those languages.

This PR updates the pipeline initialization to check Stanza’s available resources for the target language and only include the mwt processor when it is supported.

As a result, the pipeline no longer crashes for languages without MWT support, while remaining fully compatible with languages that do provide it.

Error fixed

```
ERROR: Cannot load model from D:\ebook2audiobook\models\stanza\sv\mwt\default.pt
Stanza model initialization error: Processor mwt is not known for language sv.
If you have created your own model, please specify the mwt_model_path parameter when creating the pipeline.
get_blocks() or save_json_blocks() failed! {}
```
bild

### DIFF
--- a/lib/core.py
+++ b/lib/core.py
@@ -1011,7 +1011,16 @@ INTO A NEW TRAINING MODEL. YOU CAN IMPROVE IT OR ASK TO A TRAINING MODEL EXPERT.
                             (session['device'] == devices['XPU']['proc'] and devices['XPU']['found']) or
                             (session['device'] == devices['JETSON']['proc'] and devices['JETSON']['found'])
                         ) else False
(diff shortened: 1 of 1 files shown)


ROBERT-MCDOWELL:
good catch. thanks!

ROBERT-MCDOWELL:
however keep in mind that with mwt the stanza won't work 100% for swedish.

ROBERT-MCDOWELL:
correction, mwt is not needed by swedish language... so everything is ok.

PR #1832, opened 2026-05-06
fix: Windows install fails with "13.0 was unexpected at this time" on CUDA 13.0

AddeSz:
Issue: 
On Windows with CUDA 13.0, running ebook2audiobook.cmd fails immediately after conda environment creation with:
`13.0 was unexpected at this time.`

Initial cause:
On Windows with CUDA 13.0, `check_device_info` stores a JSON string containing `"CUDA 13.0 detected via nvidia-smi"` into `DEVICE_INFO_STR`. The code row 731: `if "%DEVICE_INFO_STR%"==""` expands the JSON inline, causing CMD to choke on the bare `13.0` token with "13.0 was unexpected at this time."

Initial fix:
Replace with `if not defined DEVICE_INFO_STR` which checks existence without ever expanding the variable's value.

Initial result:
Fixed the installer from crashing and it instead continued by installing Miniforge. However, uncovered the underlying issue of `device_info_str=` being empty as seen in the following error:
```
install_device_packages() error: json.loads() could not decode device_info_str=
=============== ebook2audiobook is not correctly installed.
```

Root causes:
1. In `:check_conda`, the calls to `:check_device_info` and `:install_device_packages` were inside nested `if (...) (...)` parenthesis blocks. CMD pre-expands all `%variables%` at parse time within these blocks, so `%DEVICE_INFO_STR%` was always empty string by the time the lines executed, regardless of what `:check_device_info` had set.
2. Passing the JSON string as a command-line argument to `:install_device_packages` caused CMD to corrupt the internal double quotes of the JSON, resulting in a parse error in Python.

Fixes:

- Flattened the nested if blocks in `:check_conda` using early `exit /b` and `goto`-free logic so variables are expanded at execution time
- Changed `:install_device_packages` to receive the JSON via `DEVICE_INFO_PASS` environment variable instead of a CLI argument, so CMD's argument parser never touches the JSON content
- Applied the same `DEVICE_INFO_PASS` pattern to the `BUILD_DOCKER` call site for consistency

Result:
The `device_info_str` is correctly set and the installation proceeds successfully.

bild

Tested on: Windows 10, NVIDIA GeForce GTX 1070, CUDA 13.0 (driver-only, nvidia-smi reported). May need further testing on other configurations.

### DIFF
--- a/ebook2audiobook.cmd
+++ b/ebook2audiobook.cmd
@@ -618,31 +618,29 @@ for /f "delims=" %%i in ('where.exe python 2^>nul') do (
         )
     )
 )
-if "%CURRENT_ENV%"=="" (
-    if not exist "%SAFE_SCRIPT_DIR%\%PYTHON_ENV%" (
-        echo Creating ./python_env version %PYTHON_VERSION%…
-        call "%CONDA_HOME%\Scripts\activate.bat"
-        call conda update -n base -c conda-forge conda -y
-        call conda update --all -y
-        call conda clean --index-cache -y
-        call conda clean --packages --tarballs -y
-		call conda create --prefix "%SAFE_SCRIPT_DIR%\%PYTHON_ENV%" python=%PYTHON_VERSION% pip -y
-        ::call conda activate base
-        call conda activate "%SAFE_SCRIPT_DIR%\%PYTHON_ENV%"
-		call :check_device_info %SCRIPT_MODE%
-        if errorlevel 1 goto :failed
-		echo -------------------------------- %DEVICE_INFO_STR%
-		call :install_device_packages "%DEVICE_INFO_STR%"
-		if errorlevel 1 goto :failed
-        call :install_python_packages
-        if errorlevel 1 goto :failed
-		call conda deactivate >nul && call conda deactivate >nul
-    )
-) else (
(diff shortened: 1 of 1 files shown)


AddeSz:
Not sure if it's only on CUDA 13.0, but I marked it as such since it's the only enviornment I could test on

ROBERT-MCDOWELL:
cuda 13.0 is already in the max cuda version in conf.py. this DEVICE_INF_STR empty was a bug and fixed already.
there was a call of check_device_info() python function missing. thanks

ROBERT-MCDOWELL:
btw the fix will be available to the next coming release.

webmaxru:
For anyone landing here from a search: the deeper "empty `device_info_str`" part flagged in this PR was indeed resolved by the later refactor (flat `:provision_env` + passing the value through the environment).

The one remaining piece is the `if "%DEVICE_INFO_STR%"=="" (` line itself — it still expands the JSON inline, so on current `main` it crashes on newer CUDA with `13.x was unexpected at this time.` right after `conda create`.

I opened #1933 with just that minimal one-line guard (`if not defined DEVICE_INFO_STR`, which never expands the value). Thanks @AddeSz for the original diagnosis here.

PR #1855, opened 2026-05-08
Fixes multiple bugs with XTTS distortions. Added new sentence splitting logic.

### LINKED ISSUE #6: [feature]: Optimize sentence splitting for improved XTTSv2 prosody
Adds an optional sentence splitting strategy optimized for XTTSv2's
autoregressive decoder. Activated with --optimize_sentence_splitting.

The new splitter (lib/sentence_splitter.py) addresses two known
artefacts that appear when text is split at arbitrary boundaries:

1. Pronunciation instability - the model cannot predict natural stress
   when a clause is split mid-sentence.
2. Speed variation / pacing drift - the decoder compensates for missing
   context by slowing down or speeding up.

Strategy:
- Merge consecutive short hard-punctuation segments so clauses that
  belong together are fed as a single unit.
- When a chunk exceeds max_chars (75% of language max), split at the
  MIDPOINT of the nearest punctuation in priority order:
    .!? > : > ; > ,
- Splitting at the midpoint keeps both halves balanced, which keeps
  the model in its optimal operating range throughout the book.

Default behaviour (without the flag) is unchanged.

### LINKED ISSUE #7: [bug]: Follow-up #1791: [break]/[pause] token appears at start of sentence instead of end of preceding sentence
Follow-up #1791

When converting an EPUB, `[break]` and `[pause]` tokens marking paragraph
boundaries are displayed at the start of the next sentence instead of at
the end of the preceding sentence.

**Root cause:**

`normalize_text()` inserts a trailing space after hard punctuation on the
already-escaped text, placing it between the punctuation and the SML escape
char (`\uE000+`):

    "packen.\uE000Brauchst"  →  "packen. \uE000Brauchst"

After `restore_sml()` this produces `"packen. [break]Brauchst"` in the JSON.
`get_sentences()` then splits at the period (lookahead sees the space) and
the escaped token ends up at the start of the next sentence.

While the TTS engine correctly strips SML tags before inference, the wrong
placement produces incorrect pause timing and misleading progress output.

**Fix — three changes in `lib/core.py`:**

1. `normalize_text()` — place the space after the SML escape char, not before it
2. `get_sentences()` PASS 1 — consume trailing SML escape chars in `hard_pattern`
3. `get_sentences()` PASS 4 — don't merge backward across a paragraph boundary

### LINKED ISSUE #8: [bug]: Follow-up DrewThomasson#1791: distortion at end of sentence caused by leading space in dot replacement
Follow-up DrewThomasson#1791
When xtts converts a sentence part for inference, it replaces `.` with ` ;\n`:

```python
part = part.replace('.', ' ;\n')
```

The leading space before `;` causes the TTS model to generate an audible
distortion or artifact at the **end** of the sentence, analogous to the
`[break]` token distortion at sentence start (issue #1792).

**Fix:** Remove the leading space so the replacement is `;\n` instead of ` ;\n`.

Tested on: v26.4.30, Windows 11, CUDA 13.1, German language, XTTSv2

### LINKED ISSUE #11: [bug]: XTTSv2 produces audio loop artifacts results in horrible distortions.
XTTS generates audio auto regressively—token by token. If the model "stalls" at a punctuation mark, it repeatedly generates the same audio token. This produces an exactly periodic waveform with a constant lag across many windows looking like this wave form:

Image

The same problem sound different in many cases, because the model gets stuck at different points in different sentences:

Sometimes on a vowel → sounds like humming
Sometimes on a consonant transition → sounds like a crackling or electrical noise
Sometimes on an inhale token → sounds like static

It is the same bug (token repetition), but the repeated token sounds completely different depending on the context.

Strategie:

The problem should be solved by dividing the audio signal into short windows and determining the dominant period length for each window using auto correlation. Subsequently, it should be checked whether this period length remains nearly constant across several consecutive windows. An unusually long sequence of identical values ​​is identified as an indication of a loop artifact.

### LINKED ISSUE #1800: [feature]: add --optimize_sentence_splitting for improved XTTSv2 prosody
Adds an optional sentence splitting strategy optimized for XTTSv2's
autoregressive decoder. Activated with --optimize_sentence_splitting.

The new splitter (lib/sentence_splitter.py) addresses two known
artefacts that appear when text is split at arbitrary boundaries:

1. Pronunciation instability - the model cannot predict natural stress
   when a clause is split mid-sentence.
2. Speed variation / pacing drift - the decoder compensates for missing
   context by slowing down or speeding up.

Strategy:
- Merge consecutive short hard-punctuation segments so clauses that
  belong together are fed as a single unit.
- When a chunk exceeds max_chars (75% of language max), split at the
  MIDPOINT of the nearest punctuation in priority order:
    .!? > : > ; > ,
- Splitting at the midpoint keeps both halves balanced, which keeps
  the model in its optimal operating range throughout the book.

Default behaviour (without the flag) is unchanged.

thekk1:
## Fixes thekk1#6 and thekk1#11: Optimized sentence splitting for xtts. Fix no pause at non h1,2,3 formated chapters. Fix XTTS looping artifacts.
- Fixes DrewThomasson#1800 and additional bugs.
- Added a method for detecting typical XTTS looping artifacts larger than 700ms. Fixes thekk1#11
- Stable times for [break] and [pause].
- Set first and last sentence at the chapter to [pause] instead of break if there is just a [break].
- Added scene break patterns to add [pause] at change of scene in the book.
- Filtering HTML tags for text formatting that leads to incorrect line breaks and disrupts the conversational context.
- New text splitting method for way better speech context for the model (tested with xtts):

     1. Split chapter in paragraphs 
     2. Split too large paragraphs from the middle to the outside at hard punctuation.
     3. Split too large leftover sentences from the middle to the outside at soft punctuation.
     4. Split too large leftover sentences parts from the middle to the outside at white spaces.
     5. Merge short sentences up to 20% of max_chars.

- Using 90% of possible max chars for a text block instead of 50% resulting in better speech context. There are not significant more or less hallucinations in larger text blocks. 

## fix thekk1#7: move [break]/[pause] token to end of preceding sentence after hard punctuation split
Follow-up DrewThomasson#1791

normalize_text() inserts a trailing space after hard punctuation on the escaped text, placing it between the punctuation and the SML escape char (e.g. "packen. \uE000" instead of "packen.\uE000 "). This causes the token to appear at the start of the next sentence, producing incorrect pause timing and misleading progress output.

Fix 1 - normalize_text(): place the space after the SML escape char.

Fix 2 - get_sentences() PASS 1: consume trailing SML escape chars in hard_pattern so they stay with the preceding sentence.

Fix 3 - get_sentences() PASS 4: don't merge a short sentence backward across a paragraph boundary (sentence ending with SML escape char).

## fix thekk1#8: remove leading space in dot replacement to prevent TTS distortion at sentence end
fixes thekk1#8:

part.replace('.', ' ;\n') causes the TTS model to generate an audible
distortion at the end of sentences, analogous to the [break] token
distortion at sentence start (#1792).

The leading space before ';' is unnecessary and causes the model to
render an artifact before the sentence boundary marker.

Fix: replace with ';\n' (no leading space).


ROBERT-MCDOWELL:
provide an audio proof in the 20 main languages this PR fixes what the A.I. is pretending to for xttsv2. Provide also proof this PR does not affect fairseq, vits, tacotron2, yourtts, bark, tortoise, glowtts.

thekk1:
I am certain that you are fully aware that this will not happen—and I do not see these requirements in any other pull request for this project, either. You are welcome to try it out in your own test environment.
Apart from that, I speak neither Mandarin nor Hindi,
so I couldn't judge at all whether that would be appropriate or not.
If you can do that, wouldn't it be better if you did it yourself?
This is merely a suggestion for improving your project.
If you would prefer not to do so, that is perfectly fine with me.

PR #1873, opened 2026-05-18
Add English to Spanish translation test in workflow

DrewThomasson:
Added a test for translating English to Spanish using YOURTTS.


DrewThomasson:
@ROBERT-MCDOWELL 
Merge 👍 ?

ROBERT-MCDOWELL:
LGTM. could you also create another with eng pivot test like --language deu --translate jpn for example, which is usually not direct and needs eng pivot in between.

DrewThomasson:
so that looks good?

DrewThomasson:
or should I make the hungarian OCR test also translate to another language

If that's possible that would do a more tests at the same time

Thus lowering testing time a little

But then again these translate test addition should only add an additional...1-2 minutes extra at the most

ROBERT-MCDOWELL:
oh yes why not!

DrewThomasson:
Ok so I was able to condense the english to spanish translation

DrewThomasson:
So that looks good?

ROBERT-MCDOWELL:
yup!

PR #1875, opened 2026-05-21
Add voice shorthand test in E2A-Test.yml

DrewThomasson:


DrewThomasson:
@ROBERT-MCDOWELL 

The forgot about the shorthand for built in voices test

Am I correct to think that built-in voices use can use 

```
--voice "voices/eng/elder/male/DavidAttenborough.wav"
```

and 

```
--voice DavidAttenborough
```

should both do the exact same thing?

ROBERT-MCDOWELL:
for now there is no shorthand for --voice

DrewThomasson:
ah, thanks that makes more sense 👍 

:)

PR #1902, opened 2026-05-27
add qwen3tts mac and cpu not working correctly need to fix unknown is…

DrewThomasson:
add qwen3tts mac and cpu not working correctly need to fix unknown if cuda works or not


DrewThomasson:
This is running into issues of infinit generation loops on MPS and CPU leading to hours of nonsense generated.

Unknown how to fix at the moment

DrewThomasson:
do not merge as this is 100% broken on CPU/MPS right now

elight:
I think I know this one! I brute forced Claude into adding support on my fork last night.

here's the trick: at 0.5s of silence passing at the end of your voice sample. The model seems to otherwise hang its hat on the last phoneme, making the most hilarious and useless dribble imaginable.

Also, as an nvidia owner, I immediately turned around and had Claude replace qwen with faster-qwen3-tts. Qwen3 TTS, out of the box, seems maybe 33-50% perf of xtts v2. Faster is a drop-in replacement that is Nvidia only but uses all available GPU power. Performance is equal to or exceeds the current xtts v2 pipeline.

also got voice cloning working. Desperately needed Patrick Stewart narration.

Sadly, I'm a relative novice at Python (decades in other languages...) so I let Claude do all of the work with me guiding it.

Claude also seemed to introduce some bugs around UI statefulness—or it uncovered one. Not sure. 

Brilliant fucking work, this app, by the way!

Aside: I played with E2A-SML last night only to realize that I need more voice for it to use and then also need a way to have a way for voice cloning to fill out the voice roster. But that's a lot of effort on the user side too.

i hesitate to PR as I've done nothing to automate tests. 

Https://github.com/elight/ebook2audiobook

DrewThomasson:
So you only tested this on NVIDIA hardware?

And did you have the loop issue when you tried it without any code change on the Nvidia hardware as well?

DrewThomasson:
closing this as its not ready yet, 

Perhaps later we'll try to add this

elight:
It uses faster-qwen3-tts. That's specifically for Nvidia. However, it's a drop in replacement for qwen3-tts which I did try and worked. The faster version is just that.

sorry for the slow reply. Mail forwarding issues.

yes, tested only in Nvidia as that is all that I own.

PR #1911, opened 2026-06-09
Fix Windows cmd installer checks

lonsdaleite:
Ensure newly installed Miniforge is available in PATH before checking for conda. 
Also avoid expanding DEVICE_INFO_STR inside an IF block, which can break Windows batch parsing when device info contains special characters such as parentheses or CUDA Toolkit-related output.


ROBERT-MCDOWELL:
thanks!

PR #1913, opened 2026-06-09
Fix Windows native setup issues with Python 3.12 dependencies, UniDic SSL download failures, and BF16 numpy conversion on RTX 50xx / Blackwell GPUs.

lonsdaleite:
Changes:
- Use Python 3.11 for the Windows native environment to avoid incompatible Python 3.12 dependency resolution.
- Make UniDic download robust against Windows SSL certificate store ASN1 errors.
- Add BF16 tensor-to-numpy compatibility handling for Blackwell GPUs.


ROBERT-MCDOWELL:
- e2a does not use BF16 at all
- python 3.12 works perfectly on windows
- there is no issue to download unidic dictionnary as well as other model downloads.

Issue #1870, opened 2026-05-12
Compatibility on 5090 with CUDA 13.2 - 13.2


Neltherion:
**To report an issue here, thanks to respect strictly the process 
**or this issue will become too big and unreadable.

**IMPORTANT: DON'T COPY/PASTE ANY A.I. SOLUTION OR YOUR ISSUE WILL BE IMMEDIATELY CLOSED WITHOUT NOTICE**

1)**Operating System:**
 - Windows

2)**Hardware and Model**
 - CUDA
 - Nvidia 5090 Laptop

3)**Script Mode**
- Native
 
4)**Full Terminal log As Attached File**
Detected incomplete python_env — removing and recreating...
Creating ./python_env with python 3.12...
Channels:

* conda-forge
  Platform: win-64
  Collecting package metadata (repodata.json): done
  Solving environment: done

Package Plan

environment location: D:\AudioBook\ebook2audiobook\python_env

added / updated specs:

* pip
* python=3.12

The following NEW packages will be INSTALLED:

bzip2              conda-forge/win-64::bzip2-1.0.8-h0ad9c76_9
ca-certificates    conda-forge/noarch::ca-certificates-2026.4.22-h4c7d964_0
libexpat           conda-forge/win-64::libexpat-2.8.0-hac47afa_0
libffi             conda-forge/win-64::libffi-3.5.2-h3d046cb_0
liblzma            conda-forge/win-64::liblzma-5.8.3-hfd05255_0
libsqlite          conda-forge/win-64::libsqlite-3.53.1-hf5d6505_0
libzlib            conda-forge/win-64::libzlib-1.3.2-hfd05255_2
openssl            conda-forge/win-64::openssl-3.6.2-hf411b9b_0
packaging          conda-forge/noarch::packaging-26.2-pyhc364b38_0
pip                conda-forge/noarch::pip-26.1.1-pyh8b19718_0
python             conda-forge/win-64::python-3.12.13-h0159041_0_cpython
setuptools         conda-forge/noarch::setuptools-82.0.1-pyh332efcf_0
tk                 conda-forge/win-64::tk-8.6.13-h6ed50ae_3
tzdata             conda-forge/noarch::tzdata-2025c-hc9c84f9_1
ucrt               conda-forge/win-64::ucrt-10.0.26100.0-h57928b3_0
vc                 conda-forge/win-64::vc-14.3-h41ae7f8_34
[...]

PR #1933, opened 2026-06-16
fix: Windows installer crash "13.x was unexpected at this time" caused by JSON in DEVICE_INFO_STR

webmaxru:
N.B. This PR was co-created with AI and fully validated by me (human)

## Summary
On Windows native install, `ebook2audiobook.cmd` crashes immediately after the conda environment is created with an error like:

    13.1 was unexpected at this time.

(the number matches the detected CUDA version — reported as 13.0, 13.1, 13.2, etc.)

## Root cause
`:check_device_info` stores the JSON returned by `DeviceInstaller.check_device_info()` into `DEVICE_INFO_STR`, then tests it with:

```bat
if "%DEVICE_INFO_STR%"=="" (
```

The JSON contains embedded double quotes (e.g. the detected CUDA version inside the `"note"` field). When CMD expands `%DEVICE_INFO_STR%` inline, those embedded quotes break quote pairing and a bare token such as `13.1` gets parsed as a command, aborting the installer right after `conda create`.

## Fix
Use `if not defined`, which checks whether the variable is set **without expanding its value**, so embedded quotes/special characters can't break parsing. Also pre-clear the variable so the empty case is still detected (`for /f` skips blank output, which would otherwise leave a stale value):

```bat
:check_device_info
set "ARG=%~1"
set "DEVICE_INFO_STR="
for /f "delims=" %%I in ('python -c "..."') do set "DEVICE_INFO_STR=%%I"
if not defined DEVICE_INFO_STR (
    echo DEVICE_INFO_STR is empty
    exit /b 1
)
```

## Reproduction (minimal)
```bat
@echo off
set "DEVICE_INFO_STR={"name": "cuda", "tag": "cu131", "note": "cuda 13.1 detected"}"
if "%DEVICE_INFO_STR%"=="" ( echo empty ) else ( echo not-empty )
```
→ prints `13.1 was unexpected at this time.`
Switching to `if not defined DEVICE_INFO_STR` prints `not-empty` and the script continues.

## Testing
- Reproduced the crash and verified the fix on Windows (CMD).
- Confirmed the empty-output path still reports `DEVICE_INFO_STR is empty` and exits 1.

## Relationship to existing reports
- Fixes the still-open #1870 (CUDA 13.2 on RTX 5090, Windows / Native).
- #1832 previously flagged the same line. The later refactor resolved the deeper "empty `device_info_str`" symptom (flat `:provision_env` + passing via the environment), but the `if "%DEVICE_INFO_STR%"=="" (` line itself was never changed, so `main` still crashes. This PR is just that remaining one-line guard.

## Out of scope (noted for follow-up)
`:build_docker_image "!DEVICE_INFO_STR!"` (~line 1020) passes the JSON as a quoted CLI argument and has the same embedded-quote fragility under `--script_mode build_docker`. Left untouched to keep this fix minimal.

### DIFF
--- a/ebook2audiobook.cmd
+++ b/ebook2audiobook.cmd
@@ -759,8 +759,9 @@ exit /b 0
 
 :check_device_info
 set "ARG=%~1"
+set "DEVICE_INFO_STR="
 for /f "delims=" %%I in ('python -c "import sys; from lib.classes.device_installer import DeviceInstaller as D; r=D().check_device_info(sys.argv[1]); print(r if r else '')" "%ARG%"') do set "DEVICE_INFO_STR=%%I"
-if "%DEVICE_INFO_STR%"=="" (
+if not defined DEVICE_INFO_STR (
 	echo DEVICE_INFO_STR is empty
 	exit /b 1
 )
--- a/ebook2audiobook.cmd
+++ b/ebook2audiobook.cmd
@@ -760,7 +760,7 @@ exit /b 0
 :check_device_info
 set "ARG=%~1"
 set "DEVICE_INFO_STR="
-for /f "delims=" %%I in ('python -c "import sys; from lib.classes.device_installer import DeviceInstaller as D; r=D().check_device_info(sys.argv[1]); print(r if r else '')" "%ARG%"') do set "DEVICE_INFO_STR=%%I"
+for /f "delims=" %%I in ('python -c "import sys; from lib.classes.device_installer import DeviceInstaller as D; print(D().check_device_info(sys.argv[1]))" "%ARG%"') do set "DEVICE_INFO_STR=%%I"
 if not defined DEVICE_INFO_STR (
 	echo DEVICE_INFO_STR is empty
 	exit /b 1


ROBERT-MCDOWELL:
the python function always returns a value or empty, never false or none so your PR is wrong.
btw what's your crash log to make such kind of PR?

webmaxru:
Thanks for reviewing. Just sharing my personal experience: the latest committed version caused the described error but the suggested fix solved it.

webmaxru:
And you're right - `check_device_info()` only ever returns a JSON string or `''`, never `None`/`False`. No disagreement there.

I think the crash is actually a level lower than the Python, down in CMD's own parsing. On the **success** path, `if "%DEVICE_INFO_STR%"=="" (` expands the whole JSON inline *before* the comparison runs, and my real value happens to contain characters CMD treats as syntax - the embedded quotes plus `>`, `(`, `)`, `;`:

```
"note": "CUDA 13.1 (from nvidia-smi) > tested max 13.0.0; using cu130 torch build."
```

so it falls over with `13.1 was unexpected at this time.` right after `conda create`. The slightly counterintuitive part is that it's the *non-empty* return that trips it, not an empty/missing one.

`if not defined DEVICE_INFO_STR` just sidesteps that by checking the variable by name without ever expanding its contents. It still catches the empty case too, since `print('')` is a blank line that `for /f` skips, and the added `set "DEVICE_INFO_STR="` leaves the variable undefined.

If it helps, here's a tiny repro on current `main` 

```bat
set "DEVICE_INFO_STR={"name":"cuda","note":"CUDA 13.1 (from nvidia-smi) > x; y"}"
if "%DEVICE_INFO_STR%"=="" ( echo empty ) else ( echo ok )
```

That prints `13.1 was unexpected at this time.`; swapping in `if not defined DEVICE_INFO_STR` prints `ok`.

More than happy to tweak the PR however fits the codebase best - and thanks for maintaining this, really useful project.

webmaxru:
Attaching relevant parts of the log. It's a Windows native install on a CUDA box (the run that drove the PR). Here's the tail - it dies right after `conda create` finishes, at the `:check_device_info` step, before any pip/torch install:

```
Detected incomplete python_env — removing and recreating...
Creating ./python_env with python 3.12...
Channels:
 - conda-forge
Platform: win-64
Collecting package metadata (repodata.json): done
Solving environment: done
[… full conda package plan / install output …]
Preparing transaction: done
Verifying transaction: done
Executing transaction: done
#
# To activate this environment, use
#
#     $ conda activate <repo>\python_env
#
# To deactivate an active environment, use
#
#     $ conda deactivate

13.1 was unexpected at this time.
```

Exit code 1, and nothing after that line runs.

The reason it's hard to spot from the Python side is that the trigger is the **value** written to .device_info.json on my machine:

```json
{"name": "cuda", "os": "win", "arch": "amd64", "pyvenv": [3, 12], "tag": "cu130", "note": "CUDA 13.1 (from nvidia-smi) > tested max 13.0.0; using cu130 torch build."}
```

That `note` carries `13.1`, plus `>`, `(`, `)` and `;` — all CMD metacharacters — so when `if "%DEVICE_INFO_STR%"=="" (` expands it inline, the parser breaks on the bare `13.1`.

If you'd like to see it without a full install, this 2-liner reproduces the exact error on current `main`:

```bat
set "DEVICE_INFO_STR={"name":"cuda","note":"CUDA 13.1 (from nvidia-smi) > tested max 13.0.0; using cu130"}"
if "%DEVICE_INFO_STR%"=="" ( echo empty ) else ( echo ok )
```

→ `13.1 was unexpected at this time.` — and with `if not defined DEVICE_INFO_STR` it prints `ok`.

Happy to attach the full unedited log if that's useful.

ROBERT-MCDOWELL:
cuda version never has any special chars, the TAG is always cuxxx so cu130 or cu131 etc...
the error message is not right. the reason is elsewhere. for that I really need the full install log to see what it installed and what not. also did you have already a conda installed before other than Miniforge3?

ROBERT-MCDOWELL:
I need more info from your CUDA and environment. send an invite on discord as the solution can take more tests.

ROBERT-MCDOWELL:
ok gotcha it just need to encode the "note" value since there is < and () causing issues with batch.

ROBERT-MCDOWELL:
[device_installer.py](https://github.com/user-attachments/files/29022485/device_installer.py)
could you replace the file by this one and try again?

webmaxru:
Did not help... Log is:
```
Please update conda by running

    $ conda update -n base -c conda-forge conda



## Package Plan ##

  environment location: C:\Users\masalnik\Downloads\projects\ebook2audiobook\python_env

  added / updated specs:
    - pip
    - python=3.12


The following NEW packages will be INSTALLED:

  bzip2              conda-forge/win-64::bzip2-1.0.8-h0ad9c76_9 
  ca-certificates    conda-forge/noarch::ca-certificates-2026.5.20-h4c7d964_0 
  libexpat           conda-forge/win-64::libexpat-2.8.1-hac47afa_1 
  libffi             conda-forge/win-64::libffi-3.5.2-h3d046cb_0 
  liblzma            conda-forge/win-64::liblzma-5.8.3-hfd05255_0 
  libsqlite          conda-forge/win-64::libsqlite-3.53.2-hf5d6505_0 
  libzlib            conda-forge/win-64::libzlib-1.3.2-hfd05255_2 
  openssl            conda-forge/win-64::openssl-3.6.3-hf411b9b_0 
  packaging          conda-forge/noarch::packaging-26.2-pyhc364b38_0 
  pip                conda-forge/noarch::pip-26.1.2-pyh8b19718_0 
  python             conda-forge/win-64::python-3.12.13-h0159041_0_cpython 
  setuptools         conda-forge/noarch::setuptools-82.0.1-pyh332efcf_0 
  tk                 conda-forge/win-64::tk-8.6.13-h6ed50ae_3 
  tzdata             conda-forge/noarch::tzdata-2025c-hc9c84f9_1 
  ucrt               conda-forge/win-64::ucrt-10.0.26100.0-h57928b3_0 
  vc                 conda-forge/win-64::vc-14.5-h1b7c187_39 
  vc14_runtime       conda-forge/win-64::vc14_runtime-14.51.36231-h1b9f54f_39 
  vcomp14            conda-forge/win-64::vcomp14-14.51.36231-h1b9f54f_39 
  wheel              conda-forge/noarch::wheel-0.47.0-pyhd8ed1ab_0 



Downloading and Extracting Packages:

Preparing transaction: done
Verifying transaction: done
Executing transaction: done
#
# To activate this environment, use
#
#     $ conda activate C:\Users\masalnik\Downloads\projects\ebook2audiobook\python_env
#
# To deactivate an active environment, use
#
#     $ conda deactivate

13.1 was unexpected at this time.
```

generated `.device_info.json`:
```
{"name": "cuda", "os": "win", "arch": "amd64", "pyvenv": [3, 12], "tag": "cu130", "note": "CUDA 13.1 but tested max 13.0.0 so using torch for cu130"}
```

ROBERT-MCDOWELL:
ok so the DEVICE_INFO_STR is not involved in this issue but more some conda sub commands.
searching on it...

ROBERT-MCDOWELL:
ok replace the cmd by this one and tell me.
[ebook2audiobook.cmd.txt](https://github.com/user-attachments/files/29027344/ebook2audiobook.cmd.txt)

webmaxru:
Thanks a lot for the updated cmd! I tested it with a clean install (removed `python_env` + `.provisioned` so `check_device_info` actually re-runs): it sails past the point that used to fail and goes straight into the PyTorch install — no more `13.x was unexpected at this time`.

I've updated this PR to use your simpler `print(D().check_device_info(...))` one-liner so it matches your version. The key bit that resolves the crash is the `if not defined DEVICE_INFO_STR` guard (plus pre-clearing the variable) — that way the JSON value never gets expanded into the comparison, so the `>` / `()` / quotes in the `note` can't break CMD parsing.

Really appreciate you taking the time to dig into this and for the cleaner version!

ROBERT-MCDOWELL:
thanks!

PR #1938, opened 2026-06-19
Skip voice conversion when voice file matches target language

framallo:
## Problem

When using a voice file that's already in the target language directory (e.g., `voices/spa/.../voice.wav` with `--language spa`), `_check_xtts_builtin_speakers()` still runs the expensive voice conversion and normalize on every sentence. This causes a ~10x slowdown for non-English audiobooks using native voices.

## Root cause

The existing language check has two issues:

1. **`os.path.exists()` gate**: the early return was nested inside `if os.path.exists(current_voice)`, which fails when the working directory doesn't match the relative voice path — so the skip never fires.
2. **Missing ISO-639-1 fallback**: some voice directories use the two-letter code (`es`) rather than three-letter (`spa`). The check only tested `self.language` (ISO-639-3).

## Fix

- Add `lang_iso1` check so both `spa` and `es` in the voice path trigger the skip
- Remove the nested `os.path.exists()` gate so the early return works regardless of CWD
- 7 lines changed, 3 removed — minimal, backward-compatible

## Reproduction

```bash
# Place any Spanish .wav voice in voices/spa/adult/male/
# Run with --language spa --tts_engine xtts --voice voices/spa/adult/male/MyVoice.wav
# Without fix: "Converting builtin eng voice to spa" on every sentence
# With fix: conversion skipped, voice used directly
```

### DIFF
--- a/lib/classes/tts_engines/common/utils.py
+++ b/lib/classes/tts_engines/common/utils.py
@@ -464,9 +464,13 @@ class TTSUtils:
             import numpy as np
             from huggingface_hub import hf_hub_download
             voice_parts = Path(current_voice).parts
-            if (self.language in voice_parts or speaker in default_engine_settings[TTS_ENGINES['BARK']]['voices'] or self.language == 'eng'):
-                if os.path.exists(current_voice):
-                    return current_voice
+            # Skip conversion if the voice already matches the target language
+            lang_iso1 = getattr(self, 'language_iso1', '')
+            if (self.language in voice_parts or
+                lang_iso1 in voice_parts or
+                speaker in default_engine_settings[TTS_ENGINES['BARK']]['voices'] or
+                self.language == 'eng'):
+                return current_voice
             xtts = TTS_ENGINES['XTTS']
             if self.language in default_engine_settings[xtts].get('languages', {}):
                 default_text_file = os.path.join(voices_dir, self.language, 'default.txt')


ROBERT-MCDOWELL:
this is already coded. provide an example I can reproduce

framallo:
## Reproduction

### Setup

1. Place any Spanish `.wav` voice file (6–15 seconds of Spanish speech) under `voices/spa/adult/male/`:

```bash
# e.g. download or record a Spanish voice sample
cp /path/to/spanish_voice.wav voices/spa/adult/male/MySpanishVoice.wav
```

2. Create a minimal test EPUB:

```bash
mkdir -p /tmp/repro && cat > /tmp/repro/test.md << 'EOF'
# Capítulo 1

El viejo se sentó en el banco de la plaza. Era el mismo banco de siempre.
La plaza no había cambiado. El pueblo tampoco.

Tenía setenta y cinco años y se acordaba de todo.
EOF
pandoc /tmp/repro/test.md -o /tmp/repro/test.epub --to=epub3
```

3. Run:

```bash
./ebook2audiobook.command --headless \
    --ebook /tmp/repro/test.epub \
    --language spa \
    --tts_engine XTTSv2 \
    --voice voices/spa/adult/male/MySpanishVoice.wav \
    --output_format m4b \
    --output_dir /tmp/repro/output
```

### Expected behavior

The voice is already in `voices/spa/` — conversion should be **skipped**. The existing check at `_check_xtts_builtin_speakers` (line ~467 in `utils.py`) should catch this:

```python
if (self.language in voice_parts or ...):
    return current_voice
```

### Actual behavior

Every sentence triggers:

```
Converting builtin eng voice to spa…
Normalize completed!
```

This causes a ~10x slowdown. Here's actual output from a full book (2910 sentences):

```
4.47%: : 129/2910  : Una era del Gran Caniche…
4.47%: : 130/2910 Converting builtin eng voice to spa…
Normalize completed!
5.15%: : 149/2910  : Las Hojas habian empezado a perder su verde.
5.15%: : 150/2910 Converting builtin eng voice to spa…
Normalize completed!
5.19%: : 150/2910  : Antes eran de un verde fuerte, brillante.
5.19%: : 151/2910 Converting builtin eng voice to spa…
```

The conversion runs on **every single sentence**, not just once.

### Why

The fix in this PR changes `self.language` → `self.session['language']` and adds an `iso1` fallback check, which makes the language-matching logic actually work when called from `_set_voice()` during per-sentence processing.

ROBERT-MCDOWELL:
could you share the MySpanishVoice.wav you are using? also which version are you using?
please provide the full terminal conversion log.

ROBERT-MCDOWELL:
the folder ./voices/spa is only for official voices marked as "builtin" for all TTS engines in E2A.
you can put your own voices manually into it, but E2A will consider it as a private voice so it will be normalized, optimized and converted to the right wav specs once inside the current ./voices/__sessions/[your current session] folder.
Then, if you wish to reuse it but with the same original path which is ./voice/spa/adult/male/MySpanishVoice.wav so you must use the same session used to convert in the previous conversion like --session xxxxxxx-xxxx-xxx-xxxx.
if  you don't use --session, then E2A will consider the request with a new session and will convert it again in the new session folder.
Now indeed we can see the logic differently, consider users to put themselves their own voices into voices and run it directly, but for that it needs absolutely an audio nomalized and optimized voice.
Another logic would be you share all the spanish voices you woulld like it as official so we will do the audio job needed and reference them as official so E2A will understand there is no need to convert and optimize them.

for now just use the same --session

PR #1941, opened 2026-06-20
feat: content-addressed TTS cache + machine-readable progress/ETA (fast incremental re-renders)

framallo:
Two related changes that make iterating on a book far less painful: a content cache so small edits don't re-render the whole audiobook, and a machine-readable progress file with an accurate ETA. The second builds on the first, so they're in one PR.

---

### Problem

When you make a small edit to a book and re-convert, the whole audiobook re-renders from scratch, even though only a few sentences changed.

The existing per-sentence audio reuse only helps when **resuming an unchanged file**:

- A run without `--session` gets a fresh random session dir, so there is nothing to reuse.
- Editing the ebook changes its checksum, which deletes `blocks_saved` / `blocks_orig` / the progress DB and **regenerates random `uuid4` block ids**, orphaning the position-keyed sentence audio at `chapters/sentences/<block_id>/<j>.flac`.

So in practice any edit forces a full re-render.

### What this does

Each rendered sentence is stored at `tmp/tts_cache/<sha1>.<ext>` keyed by its content:

`engine + fine_tuned + language + voice(content) + engine params (e.g. xtts_*) + sentence text`

- **Cache hit:** hardlink the existing audio in, no inference.
- **Cache miss:** render, then populate the cache.

Because the key is the *content* (not session id, block id, or sentence position) and the cache lives **outside `process_dir`**, it survives session churn, edits, and reordering. Editing one paragraph re-renders only that paragraph; everything else, even a brand-new session with no `--session`, is served from cache.

Notes:
- **Voice is keyed by byte content, not path.** The engine copies the source voice to a per-session `voices/__sessions/voice-<id>/...` path that changes every run, so a path-based key would never hit. The processed bytes are deterministic from the source; the hash is memoized per file.
- **Engine-agnostic.** It wraps the common `convert_sentence2audio` entry point and folds every `<engine>_*` session param into the key.
- Enabled by default; disable with `EBOOK2AUDIO_TTS_CACHE=0`. Inspect keys with `EBOOK2AUDIO_CACHE_DEBUG=1`. Bump `tts_cache_version` to invalidate.

### Testing (XTTS, Spanish, CPU)

| Scenario | Result |
|---|---|
| Cold run | renders 2 paragraphs, writes 2 cache files |
| Warm run, fresh session (no `--session`) | **2 reuses, 0 TTS inferences**, cache unchanged |
| Edit one paragraph | **1 reuse + 1 inference**, cache grows by exactly 1 |

---

### Problem

The console only shows a tqdm bar (carriage-return updates, no ETA), so a script or agent driving a conversion can't read progress or estimate completion.

### What this does

A `ProgressTracker` writes a JSON snapshot to `tmp/progress/<session>.json` (and `tmp/progress/latest.json`) on every update: `status`, `percent`, `done`/`total`, rendered vs reused counts, elapsed, average render time, and ETA (seconds, human-readable, and absolute finish epoch). The path is printed as `PROGRESS_JSON: <path>` at startup. Writes are throttled (~1/s) and atomic (temp file + `os.replace`).

The ETA is **accurate because of the cache**: a pre-scan counts, before any rendering, exactly how many sentences are cache misses (must render) vs hits (reused, ~instant). So `ETA = remaining_renders × recent_avg_render_time` rather than a naive sentence count. An edit touching one paragraph reports `to_render: 1` and a small ETA even in a large book.

### Example snapshot

```json
{
  "status": "running",
  "totals":   { "sentences": 6, "to_render": 1, "cached_at_start": 5 },
  "progress": { "percent": 50.0, "done": 3, "rendered": 3, "reused": 0, "render_remaining": 3 },
  "timing":   { "eta_human": "21s", "eta_seconds": 21, "avg_render_seconds": 7.04, "eta_finish_epoch": 1781995960 }
}
```

### Testing

| Scenario | Result |
|---|---|
| Full render | ETA converged 21s, 13s, 0s, matched the 37s actual; ended `status: done` |
| Edit one paragraph | pre-scan reported `to_render: 1, cached_at_start: 5` before rendering; finished in 4s (`rendered 1, reused 5`) |

---

## Scope

One file (`lib/core.py`): cache helpers next to `block_hash`, a cache lookup/populate around the existing `convert_sentence2audio` call, a `ProgressTracker`, and the tracker calls in the conversion loop. No behavior change when the cache is disabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ https://github.com/user-attachments/assets/81c4baad-117e-4db5-ac86-efc2b7fea921
   - [Run Locally](#instructions)
     - [Launching Gradio Web Interface](#instructions)
     - [Basic Headless Usage](#basic-usage)
+    - [Incremental re-renders & progress tracking](#incremental-re-renders--progress-tracking)
     - [Headless Custom XTTS Model Usage](#example-of-custom-model-zip-upload)
     - [Help command output](#help-command-output)
   - [Run Remotely](#run-remotely)
@@ -114,6 +115,8 @@ https://github.com/user-attachments/assets/81c4baad-117e-4db5-ac86-efc2b7fea921
 - 🧩 **Optional custom model** using your own trained model (XTTSv2, VITS, FAIRSEQ, PIPER, others on request)
 - 🎛️ **Fine-tuned preset models** trained by the E2A Team<br/>
      <i>(Contact us if you need additional fine-tuned models, or if you’d like to share yours to the official preset list)</i>
+- ⚡ **Incremental re-renders** — a content-addressed cache reuses unchanged audio, so editing your book only re-renders the sentences that changed ([see below](#incremental-re-renders--progress-tracking))
+- 📈 **Machine-readable progress + ETA** — every run writes a `progress.json` with live status, percent, and an accurate ETA ([see below](#incremental-re-renders--progress-tracking))
 
 
 ##  Hardware Requirements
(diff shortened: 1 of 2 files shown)


ROBERT-MCDOWELL:
Before to create an A.I. PR, open an issue, talk as a normal human to us, expose your issue with proof of the issue, then maybe a PR will be created. this PR has absolute non sense again as this A.I. tries to change the logic of what is an ebook to audiobook logic conversion. Be sure if we blindly merge any A.I. PR without to control the code robustness, how the A.I. understands the e2a context so the project could quickly fall like a castle cards among spaghettis code.

ROBERT-MCDOWELL:
FYI, depend what you mean by small edits, text? voice? tts engine? fine tuned? etc....
E2A already manage each block (future chapter). leave untouched if nothing changed. if one param abov changed so the whole block should be reconverted to keep sync with the timestamp and the VTT (subtitle final file).

framallo:
What I meant is that the e2a process would take a lot of time after making small edits on paragraphs. Changing 1 word in 10% of paragraphs would take 90% the same time as running it from scratch. 

Anyway, I appreciate the feedback.

ROBERT-MCDOWELL:
we planned to add a sentence edit so only the sentence will be reconverted but you won't avoid the whole process of combining all sentences and chapters for the export in any case since the timestamp for the vtt file should be recalculated.
the best if you want to change a word is to do it before the conversion ;)

PR #1945, opened 2026-06-22
Fix headless Docker runs on CPU images (#1829)

### LINKED ISSUE #1829: Docker latest image fails in headless mode: USER unbound variable, sudo not found
### Describe the bug

The current Docker image `athomasson2/ebook2audiobook:latest` seems broken for the documented headless Docker usage.

A command that used to work for me now fails after pulling the latest image. The container starts `ebook2audiobook.command`, but inside the container:

- `USER` is empty/unset
- `sudo` is not installed
- the script tries to install runtime dependencies (`cmake`, `pkg-config`, `cargo`, `rust`, `wget`, etc.)
- this causes the container to fail immediately

This looks like the image is entering an installation/bootstrap path intended for native Linux usage, instead of behaving like a ready-to-run Docker image.

### To Reproduce

I run:

```bash
docker run --pull always --rm \
  -p 7860:7860 \
  -v /home/claudio/ebook2audiobook/input_folder:/app/input_folder \
  -v /home/claudio/ebook2audiobook/audiobooks:/app/audiobooks \
  athomasson2/ebook2audiobook \
  --headless \
  --ebook /app/input_folder/my_book.epub \
  --language ita
```

### Actual behavior

First error:

```text
ebook2audiobook.command: line 182: USER: unbound variable
ebook2audiobook.command: line 185: USER: unbound variable
```

If I force `USER`, for example with `-e USER=claudio`, I get:

```text
Adding claudio to group root (requires sudo)...
ebook2audiobook.command: line 192: sudo: command not found
```

If I force `USER=root`, I get:

```text
cmake is not installed.
pkg-config is not installed.
cargo is not installed.
rust is not installed.
Installing required programs. NOTE: you must have 'sudo' priviliges to install ebook2audiobook.
ebook2audiobook.command: line 530: sudo: command not found
wget is missing! trying to install it…
Cannot 'wget'. Please install 'wget' manually.
```

### Diagnostics

Inside the container:

```bash
docker run --rm --entrypoint bash athomasson2/ebook2audiobook -lc 'whoami; echo "USER=$USER"; command -v sudo || echo no-sudo'
```

Output:

```text
root
USER=
no-sudo
```

So the container is already running as root, but the script still expects `USER` to exist and later tries to use `sudo`, which is not present.

### Expected behavior

The Docker image should run headless conversion directly without requiring:
- `USER` to be preset
- `sudo` inside the container
- runtime installation of system dependencies like `wget`, `cargo`, `rust`, etc.

### Workaround

I still had an older local image and that one works:

- broken latest digest: `sha256:10237e3fdfbf112bf12439dc5c2390f5e7a1b77451fe6585651abeee63e7481e`
- older working local image ID: `2712cf63bfc4`
- older local digest: `sha256:9a8ea0dea9f37cf79f3d4f4af2ef165546818e24841956e0c1b9e05e5b85fc26`

So this seems like a regression in the current image/script path.

### Environment

- Host OS: Linux
- Docker image: `athomasson2/ebook2audiobook:latest`
- Run mode: headless Docker
- Command uses mounted folders for `/app/input_folder` and `/app/audiobooks`

erikosvaldsson:
## Fixes #1829 — headless Docker runs broken on the CPU image

Passing custom arguments to `docker run` overrides the image's default `CMD`
(`--script_mode full_docker`), which exposed four bugs that made headless CPU
runs fail. Each blocked the next, so the repro walks through all of them.

### Repro (before)
```
docker run -v ./ebooks:/app/ebooks -v ./audiobooks:/app/audiobooks \
  -v ./models:/app/models -v ./voices:/app/voices -v ./tmp:/app/tmp \
  --rm -it -p 7860:7860 athomasson2/ebook2audiobook:cpu \
  --headless --ebook /app/ebooks/book.epub --language eng
```
→ `ebook2audiobook.command: line 183: USER: unbound variable`
then, once each is fixed, in sequence:
`sudo: command not found` → missing `tools/detect_gpus.py` →
`ImportError: libcudart.so.13: cannot open shared object file`.

### Root causes & fixes
1. **`USER` unbound** under `set -u` (USER is unset in the container) — fall
   back to `id -un`.
2. **native-mode fallthrough**: dropping the default CMD left
   `SCRIPT_MODE=native`, so the script tried to reinstall the build-only deps
   the image deliberately purges (and called `sudo`, absent in the image). Mark
   the runtime image with `E2A_DOCKER_RUNTIME=1` and default to `full_docker`
   when it's set and no `--script_mode` was given.
3. **`tools/detect_gpus.py` missing**: `.dockerignore` excluded all of `tools/`,
   but `device_installer` runs that script. Narrow the ignore to keep it.
4. **`onnxruntime-gpu` on the CPU image** (pulls in the CUDA runtime): swap
   `onnxruntime-gpu` → `onnxruntime` for `cpu` targets in the bash build-path
   installer, plus the same guard in the Python device installer.

### Notes for reviewers
- **GPU builds unaffected**: the onnxruntime swap is gated on `tag == cpu`, and
  the `full_docker` default only triggers inside the prebuilt image (via the
  `E2A_DOCKER_RUNTIME` env baked in after the build step).
- There are **two** `install_python_packages` — a bash one (Docker build path)
  and a Python one (native/runtime path). Both are patched so the fix holds
  regardless of entry point.

### After
The same command runs to completion and produces the audiobook.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -12,6 +12,7 @@ audiobooks/cli/*
 audiobooks/gui/gradio/*
 audiobooks/gui/host/*
 models/*
-tools
+tools/*
+!tools/detect_gpus.py
 tmp/*
 run/*
--- a/Dockerfile
+++ b/Dockerfile
@@ -79,6 +79,10 @@ RUN find /app -type f \( -name "*.sh" -o -name "*.command" \) -exec sed -i 's/\r
 # Build dependencies via project script
 RUN ./ebook2audiobook.command --script_mode build_docker --docker_device "$DOCKER_DEVICE_STR"
 
+# Mark this as the prebuilt runtime image so the entrypoint defaults to
+# full_docker mode even when custom `docker run` args override the default CMD.
+ENV E2A_DOCKER_RUNTIME=1
+
 # Cleanup build-only packages and Rust toolchain to shrink the image
 RUN set -eux; \
     rustup self uninstall -y 2>/dev/null || true; \
--- a/ebook2audiobook.command
+++ b/ebook2audiobook.command
@@ -116,6 +116,12 @@ if [[ -n "${arguments[script_mode]+exists}" ]]; then
         echo "Error: Invalid script mode argument: ${arguments[script_mode]}"
         exit 1
     fi
+elif [[ "${E2A_DOCKER_RUNTIME:-0}" == "1" ]]; then
+    # Inside the prebuilt runtime image the build-only deps (cmake, rust, wget…)
+    # have been purged, so default to full_docker. This makes custom `docker run`
+    # args work without the user having to re-specify --script_mode full_docker
+    # (passing args to `docker run` overrides the image's default CMD).
+    SCRIPT_MODE="$FULL_DOCKER"
 fi
 
 if [[ -n "${arguments[docker_device]+exists}" ]]; then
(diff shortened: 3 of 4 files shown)


ROBERT-MCDOWELL:
your A.I. solution is a mess and start (as many A.I.) to create spaghettis code. the issue does not come from docker or the code itself, it comes from github that didn't upload on time the update adding detect_gpus.py script.
did you try the v26.6.17?

erikosvaldsson:
Yes, v26.6.17 has the same issue.

ROBERT-MCDOWELL:
btw to build a docker it's --script_mode build_docker, not --script_mode full_docker.
full_docker is an internal option value the  user should never use.

erikosvaldsson:
Fair enough, even with it the same issue persist.
build_docker is what I have used for the manual local builds.

ROBERT-MCDOWELL:
I'm working on it to find the right solution

PR #1959, opened 2026-07-02
fix(core): add next_hour substitution for polish

gucio321:
This caused python exception and stopped conversion.


ROBERT-MCDOWELL:
fixed on the new v26.7.7 thanks

PR #2007, opened 2026-07-14
fix: process all files in GUI directory mode

ShiroKSH:
## Summary

- Process every supported file selected in GUI Directory mode.
- Previously the conversion handler returned after the first successful file.

## Verification

- python -m compileall -q app.py lib
- python app.py --help
- Focused two-file regression harness exercised the real handler and verified both files were converted in order.


ROBERT-MCDOWELL:
Thanks!

PR #2029, opened 2026-07-21
fix(deps): replace legacy onnx-simplifier with onnxsim in requirements.txt

walruscraft:
The upstream project renamed its official PyPI package from onnx-simplifier to onnxsim. This has caused a multiple versions install loop.


ROBERT-MCDOWELL:
thanks!

PR #2053, opened 2026-08-09
Preserve sentence audio tails

0cwa:
Preserve natural sentence-ending audio tails while capping added silence
    at 50 ms. Add dependency-free regression tests for tail bounds and slice
    boundaries.
    
Created with GPT 5.6 Luna xhigh to fix the end of sentences being cut off.

### DIFF
--- a/lib/classes/tts_engines/common/audio.py
+++ b/lib/classes/tts_engines/common/audio.py
@@ -42,6 +42,21 @@ def detect_gender(voice_path:str)->str|None:
         print(error)
         return None
 
+NATURAL_TAIL_TARGET_SEC = 0.020
+NATURAL_TAIL_MAX_SEC = 0.050
+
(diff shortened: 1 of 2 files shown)


ROBERT-MCDOWELL:
No A.I. PR please if you don't know python code.
anyhow, each TTS engine has its own behavious at the end of a sentence, depending the punctuation, the language, the model etc... so your PR does it for a global behavior which is absolutely not the solution.
Rather, open an issue, and explain clearly what's wrong in your conversion, providing all info requested, especially the TTS engine you are using, language, model etc.... also provide the file you convert with the issue so I can reproduce it.

0cwa:
Fair, I opened it as a draft to work on in public and this was just a start (not used to such responsive maintainers 🙃 ) #2054

PR #2061, opened 2026-08-22
fix: preserve natural sentence boundaries

lucasPaulDaniele:
# Fix Sentence Segmentation for More Natural Audiobook Narration

## Problem

During ebook-to-audiobook conversion, long passages could be split at unnatural positions:

- in the middle of a clause;
- immediately after conjunctions such as `and`, `but`, or `because`;
- at a comma located much later in the text;
- between several intentionally short expressive sentences;
- across a real sentence boundary.

This could produce overly long segments and unnatural narration.

XTTS also rejects input exceeding its maximum limit of 400 tokens.

## Real-World Example

For example, this French passage could previously be split like this:

```text
Bosch régla son scanner sur la fréquence des services d'intervention du comté de Los Angeles et
entendit les capitaines des détachements de pompiers informer le poste de commandement que neuf maisons
avaient déjà été détruites dans une rue,
celles de la rue voisine se trouvant maintenant sur le chemin des flammes.
```

The first split occurs immediately after the conjunction `et`, leaving the first segment grammatically unfinished.

The passage means:

> Bosch tuned his scanner to the frequency used by the Los Angeles County emergency services and heard the fire department captains inform the command post that nine houses had already been destroyed on one street, while the houses on the neighboring street were now in the path of the flames.

A more natural segmentation is:

```text
Bosch régla son scanner sur la fréquence des services d'intervention du comté de Los Angeles et entendit les capitaines des détachements de pompiers informer le poste de commandement que neuf maisons avaient déjà été détruites dans une rue,
celles de la rue voisine se trouvant maintenant sur le chemin des flammes.
```

This keeps the complete clause together and uses the later comma as the narration break.

Another real example contains several short, intentionally expressive sentences:

```text
Il regarda l'escadrille des hélicoptères; semblables à des libellules à cette distance, ils zigzaguaient entre les nuages de fumée avant de larguer des tonnes d'eau et de retardateur de couleur rose sur les maisons et les arbres en feu. Cela lui rappela les offensives aériennes au Vietnam. Le bruit. La danse hésitante des appareils surchargés. Des masses d'eau traversaient des toits en feu, de la vapeur s'élevant aussitôt dans le ciel.
```

The English translation is:

> He watched the helicopter formation; like dragonflies at that distance, they zigzagged between the clouds of smoke before dropping tons of water and pink retardant on the burning houses and trees. It reminded him of the air offensives in Vietnam. The noise. The hesitant dance of the overloaded aircraft. Masses of water passed through burning roofs, steam rising immediately into the sky.

The sentences `Le bruit.` and `La danse hésitante...` are deliberate narrative fragments and should remain separate.

## Changes

The sentence splitter now:

1. Prioritizes real sentence endings such as `.`, `!`, and `?`.
2. Prevents commas and clause connectors from crossing real sentence boundaries.
3. Prefers intermediate punctuation such as `;`, `:`, and `,` inside a long sentence.
4. Prefers a natural comma break over an earlier clause connector when appropriate.
5. Uses clause connectors only as a fallback when punctuation is unavailable.
6. Preserves short expressive sentences instead of merging them with neighboring text.
7. Keeps linguistic connector data outside the core algorithm.
8. Splits oversized XTTS inputs before inference to avoid the 400-token limit.

## Language Configuration

Clause connectors are now stored in `lib/conf_lang.py` instead of being embedded directly in `lib/core.py`.

This makes the logic easier to maintain and allows language-specific connector lists to be added later without modifying the segmentation algorithm.

The existing fallback list is preserved to avoid changing behavior for languages that do not yet have dedicated configuration.

## Tests

Regression tests cover:

- real sentence boundaries;
- long clauses;
- long text without punctuation;
- comma preference over an earlier conjunction;
- preservation of complete final phrases;
- short expressive sentences.

Validation result:

```text
6 passed
```

## Known Limitations

This remains a heuristic splitter and does not perform complete syntactic analysis.

Some cases may still require additional language-specific rules, including:

- very long sentences without punctuation;
- abbreviations containing periods;
- complex dialogue;
- incorrectly punctuated source text;
- languages without dedicated connector configuration;
- unusual grammatical structures.

The current fallback behavior is intentionally preserved for compatibility.

### DIFF
--- a/lib/__init__.py
+++ b/lib/__init__.py
@@ -18,6 +18,7 @@ from .conf_lang import (
     language_math_phonemes, language_clock, os, punctuation_list, 
     punctuation_list_set, punctuation_split_hard, punctuation_split_hard_set,
     punctuation_split_soft, punctuation_split_soft_set, punctuation_switch,
+    default_clause_split_words, language_clause_split_words,
     specialchars_mapping, chars_remove, year_to_decades_languages,
 )
 
@@ -48,6 +49,7 @@ __all__ = [
     "language_math_phonemes", "language_clock", "os", "punctuation_list", 
     "punctuation_list_set", "punctuation_split_hard", "punctuation_split_hard_set",
     "punctuation_split_soft", "punctuation_split_soft_set", "punctuation_switch",
+    "default_clause_split_words", "language_clause_split_words",
     "specialchars_mapping", "chars_remove", "year_to_decades_languages",
     
     # from conf_models
--- a/lib/classes/tts_engines/xtts.py
+++ b/lib/classes/tts_engines/xtts.py
@@ -113,6 +113,25 @@ class XTTS(TTSUtils, TTSRegistry, name='xtts'):
             error = f'load_engine() error: {e}'
             raise RuntimeError(error) from e
 
+    def _split_xtts_text(self, text:str, language:str|None=None)->list[str]:
+        try:
+            lang = (language or self.language_iso1 or self.language or 'en').split('-')[0]
+            if not hasattr(self.engine, 'tokenizer'):
+                return [text]
+            limit = getattr(self.engine.tokenizer, 'char_limits', {}).get(lang, 250)
+            if not limit or len(text) <= limit:
+                return [text]
+            try:
+                import TTS.tts.layers.xtts.tokenizer as xtts_tokenizer
+                chunks = xtts_tokenizer.split_sentence(text, lang, limit)
+                if chunks and any(chunk.strip() for chunk in chunks):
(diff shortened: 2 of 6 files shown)


ROBERT-MCDOWELL:
open an issue if you encounter any split sentences problems. you must know that a TTS won't guess it's the continuity of a sentence when e2a has to split before the limited token so the "unnatural" voice is not avoidable when not enouigh punctuation nor punctuation at the right place exists.

ROBERT-MCDOWELL:
and please. STOP TO COPY PASTE A.I. wrong solutions.

PR #2075, opened 2026-09-06
Add Kokoro-82M and Fish-Speech (s2-pro) TTS engines

aaron-ang:
## Summary
- Adds `Kokoro-82M` as a new TTS engine (Apache-2.0, commercial-safe like the rest of this repo's engines).
- Adds `Fish-Speech` (s2-pro) as a new TTS engine, run via its own local API server (`tools/api_server.py`) rather than an in-process model load.
- Both engines follow the existing preset/registry pattern in `lib/classes/tts_engines/`.

## Licensing note
`Fish-Speech` (s2-pro) model weights are under the Fish Audio Research License — **non-commercial use only**. Commercial use requires a paid license from Fish Audio (business@fish.audio). Called out in the README, in both the engine list and the TTS-engines roadmap checklist, since it differs from every other engine here. `Kokoro-82M` has no such restriction (Apache-2.0).

Neither engine gets batched inference, unlike the Breeze engine in #2077; the reasons are in inline comments on the diff, so they don't get re-litigated later.

## Test plan
- [x] `pytest tests/tts_engines/test_kokoro.py -v` against a real model load.
- [x] `pytest tests/tts_engines/test_fish.py -v` against a live `api_server.py` instance (1 passed in 83.51s).
- [x] `--headless --ebook` conversion end-to-end for both engines on CUDA, valid `.m4b` with ffprobe-confirmed sample rate and matching VTT.
- [x] Fixed a pre-existing `UnboundLocalError` in `device_installer.py`'s `detect_device()` that crashes every native `app.py` launch on CUDA-only (no ROCm) hardware — unrelated to these engines, found while testing on real hardware.
- [ ] The narrowed `except` clauses are compile-checked and reviewed but not re-exercised by a fresh end-to-end run; they change which exceptions propagate, not the success path.


lib/classes/tts_engines/fish.py
```diff
@@ -0,0 +1,238 @@
...
+
+                    snapshot_download(
+                        repo_id=self.models[self.session["fine_tuned"]]["repo"],
+                        local_dir=checkpoint_dir,
+                    )
+                # sys.executable (not a bare 'python') so the subprocess shares this
+                # app's already-resolved torch/transformers venv.
+                proc = subprocess.Popen(
+                    [
+                        sys.executable,
```
aaron-ang, line 85: Uses sys.executable so the subprocess shares this app's already-resolved torch/transformers venv rather than whatever `python` resolves to on PATH.

lib/classes/tts_engines/fish.py
```diff
@@ -0,0 +1,238 @@
...
+                        os.path.join(FISH_SPEECH_REPO_DIR, "tools", "api_server.py"),
+                        "--listen",
+                        f"{FISH_API_HOST}:{FISH_API_PORT}",
+                        "--llama-checkpoint-path",
+                        checkpoint_dir,
+                        # Defaults to a path relative to server cwd, independent of
+                        # --llama-checkpoint-path; must be set explicitly.
+                        "--decoder-checkpoint-path",
```
aaron-ang, line 93: Fish-Speech's tools/server/api_utils.py defaults this to a path relative to the server's cwd, independent of --llama-checkpoint-path. Left unset, the DAC decoder fails to load (confirmed via a crashed run: FileNotFoundError on that relative path).

lib/classes/tts_engines/fish.py
```diff
@@ -0,0 +1,238 @@
...
+                        os.path.join(checkpoint_dir, "codec.pth"),
+                    ],
+                    cwd=FISH_SPEECH_REPO_DIR,
+                )
+                import time
+
+                # ModelManager does an unconditional warmup generation before /health
+                # returns 200; budget generously (15 min) for a cold start under load.
+                for _ in range(180):
```
aaron-ang, line 102: ModelManager.__init__ runs an unconditional warmup generation before /health returns 200 - a cold start under GPU contention took well over 120s in testing, hence the generous 15-minute budget.

lib/classes/tts_engines/fish.py
```diff
@@ -0,0 +1,238 @@
...
+                    if self._server_is_up():
+                        break
+                    time.sleep(5)
+                else:
+                    proc.terminate()
+                    try:
+                        proc.wait(timeout=10)
+                    except subprocess.TimeoutExpired:
+                        # SIGTERM is ignored mid blocking-CUDA-call; SIGKILL frees the GPU.
+                        proc.kill()
```
aaron-ang, line 112: SIGTERM was observed to be ignored while the subprocess is mid a blocking CUDA generation call; SIGKILL is needed to actually free the GPU.

lib/classes/tts_engines/fish.py
```diff
@@ -0,0 +1,238 @@
...
+                        f"http://{FISH_API_HOST}:{self.engine['port']}/v1/tts",
+                        params={"format": "msgpack"},
+                        data=ormsgpack.packb(
+                            req, option=ormsgpack.OPT_SERIALIZE_PYDANTIC
+                        ),
+                        headers={"content-type": "application/msgpack"},
+                        # Autoregression runs to its own stop token, not to len(part);
+                        # budget generously rather than a short number that only works
+                        # uncontended.
+                        timeout=600,
```
aaron-ang, line 196: Measured ~1.7s/semantic-token during generation. Per-sentence generation isn't capped short (autoregression runs to its own stop token, not to len(part)), so even a short sentence can exceed a 120s read-timeout under GPU contention.

lib/classes/tts_engines/fish.py
```diff
@@ -0,0 +1,251 @@
...
+                try:
+                    req = ServeTTSRequest(
+                        text=part,
+                        references=references,
+                        reference_id=None,
+                        format="wav",
+                        streaming=False,
+                        use_memory_cache="on",
+                    )
+                    resp = requests.post(
```
aaron-ang, line 187: Why Fish gets no batched path, unlike Breeze in #2077. It would benefit — it is autoregressive, so it has the same memory-bandwidth ceiling at batch 1 — but it needs a rewrite of vendored inference rather than a port. `generate()` in `fish_speech/models/text2semantic/inference.py` is hard-wired to batch 1: `setup_caches(max_batch_size=1)` (commented "Fixed to 1, avoid dynamic changes"), `prompt.view(1, codebook_dim, -1)`, a `seq` buffer with no batch dimension, and `decode_one_token` under `torch.compile(fullgraph=True)`.

Two cheaper routes were checked and ruled out: `group_turns_into_batches` is text packing (joins with `\n`) and yields no per-sentence audio boundaries, which this pipeline needs for resume, VTT and chapter combine; and firing concurrent HTTP requests doesn't help because the server runs one queue worker against one GPU-bound model.

aaron-ang (commented):
Inline notes on the rationale behind the less obvious choices, kept out of the PR description.

lib/classes/tts_engines/kokoro.py
```diff
@@ -0,0 +1,161 @@
...
+                    continue
+                if SML_TAG_PATTERN.fullmatch(part):
+                    success, error = self._convert_sml(part)
+                    if not success:
+                        return False, error
+                    continue
+                if not any(c.isalnum() for c in part):
+                    continue
+                try:
+                    generator = self.engine(part, voice=voice_id, speed=1)
```
aaron-ang, line 118: Why Kokoro gets no batched path: there is no batch dimension to fill. `KModel.forward(phonemes: str, ref_s, speed)` takes a single phoneme *string*, and `KPipeline.__call__` accepts a list but is a generator that processes sequentially.

The payoff would be small regardless — at 82M params and not per-frame autoregressive, it isn't memory-bandwidth bound at batch 1, which is the effect that makes batching a large win for Breeze.

lib/classes/tts_engines/fish.py
```diff
@@ -0,0 +1,251 @@
...
+
+                    audio_np, sr = sf.read(tmp_wav, dtype="float32")
+                    Path(tmp_wav).unlink(missing_ok=True)
+                    if sr != self.params["samplerate"]:
+                        audio_np = self._resample_audiodata(
+                            audio_np, sr, self.params["samplerate"]
+                        )
+                    part_tensor = self._tensor_type(audio_np).unsqueeze(0)
+                    self.audio_segments.append(part_tensor)
+                except (
```
aaron-ang, line 217: Ruff BLE001: every blind `except Exception` in this file and in `kokoro.py` is narrowed to what the guarded code can legitimately raise, so real bugs stop being swallowed alongside the failures these handlers exist to absorb. This one covers what a single part can hit — network, tmp-file I/O, decode/resample.

`fish.py`'s imports were hoisted above the `try` so these clauses can name `requests.RequestException` without a `NameError`.

lib/classes/tts_engines/fish.py
```diff
@@ -0,0 +1,251 @@
...
+                if self.session["device"]
+                in [
+                    devices["CUDA"]["proc"],
+                    devices["ROCM"]["proc"],
+                    devices["JETSON"]["proc"],
+                ]
+                else self.session["device"]
+            )
+            self.engine = self.load_engine()
+        except (KeyError, OSError, RuntimeError, ValueError) as e:
```
aaron-ang, line 43: `load_engine()` below keeps its broad `except Exception` on purpose, in both engines: it re-raises as `RuntimeError` rather than returning a failure tuple, so nothing is hidden and the original traceback is chained. This clause is the one that needed narrowing, since it is the boundary where a bug here would have been reported as a config error.

ROBERT-MCDOWELL:
requirements.txt version pkg must not be touched. also which version of e2a did you create the PR since most of your PR is already fixed.

aaron-ang:
Based on `upstream/main` @ `593a5b2c`, `VERSION.txt` `26.8.20` — `git log upstream/main ^HEAD` is 0, so not behind main. Which parts are already fixed in the next version? If Kokoro and Fish are in it, I'll close this.

On requirements.txt: `./ext/py/fish-speech` mirrors the existing `./ext/py/demucs` line. `kokoro>=0.9.4` and `ormsgpack` are plain PyPI deps for the engines — `DeviceInstaller.device_pkgs` is only for packages whose variant depends on the device. Where do you want engine deps declared instead?

ROBERT-MCDOWELL:
demucs repo has been added into e2a repo since it's not maintained anymore and since I applied patches to accept higher deps versions.

PR #2076, opened 2026-09-06
chore: stop tracking generated build artifacts

aaron-ang:
## Summary
- `ext/py/demucs/build/lib/` and both `ebook2audiobook.egg-info/` / `ext/py/demucs/demucs.egg-info/` were tracked in git — these are 100% reproducible output of `pip install`/`setup.py build`, not source, and standard practice is to gitignore them rather than commit them.
- Added `*.egg-info/` and `build/` to `.gitignore`, removed the 51 already-tracked generated files from git's index (kept on disk, still work identically — nothing is deleted from anyone's working copy on pull, just untracked going forward).
- Also gitignored `.venv/`, `voices/` (MMS/Piper voice-pack downloads), and `.device_info.json` (device_installer's runtime cache) — same category of local-only clutter, noticed at the same time.

Split out of the Kokoro/Fish-Speech PR (#2075) since it's an unrelated pre-existing repo-hygiene issue, noticed while auditing that PR's own working tree.

## Test plan
- [x] `git status` clean after the removal — no unintended files caught by the new patterns.
- [x] Verified these files aren't referenced by any import path outside their own package (they're regenerated automatically the next time anyone runs `pip install ./ext/py/demucs` or `pip install -e .`).


ROBERT-MCDOWELL:
the user must NOT run anything from ext/py. only the sh/cmd script must run it.

PR #2077, opened 2026-09-06
feat: add Breeze-TTS-2 engine

aaron-ang:
## Summary
- Adds Breeze-TTS-2 as a new TTS engine, vendored at `ext/py/breeze-tts` (same pattern as `ext/py/demucs` — no upstream `setup.py`/`pyproject.toml`, so this repo authors one and installs it via `requirements.txt`). The vendored package's internal `models` module is renamed to `breeze_models` to avoid colliding with other editable installs.
- Runs as a local API server (`breeze_infer.api`), started on first use and reused across requests. CUDA-only, no CPU inference path.
- Adds a batched inference endpoint and engine path, which is what makes full-length books practical here: **0.53–0.58 sentences/s at batch 64 versus ~0.120 unbatched** (same book, same machine, single GB10) — about 4.4x, taking one 10,153-sentence book from ~17 h to ~3.9 h of remaining wall-clock.
- Registered in `conf_models`, smoke test at `tests/tts_engines/test_breeze.py`.

Also fixes four real bugs found getting this hardware-tested end-to-end: inconsistent narrator voice per sentence, a silent `transformers>=5.0.0` upgrade that broke model load, missing `configs/*.json` packaging, and a `TypeError` in `convert_ebook()`'s resume branch. Details are in inline comments on the diff.

## Test plan
- [x] `pytest tests/tts_engines/test_breeze.py` against an isolated server instance.
- [x] Three sentences via independent `convert()` calls, narrator identity consistent across all three.
- [x] End-to-end CLI conversion (`app.py --headless --tts_engine breeze --device cuda`) producing `.m4b` + `.vtt`.
- [x] Batch endpoint verified with a standalone client: `X-Segment-Bytes` present, segment count equal to text count, sizes summing to body length, each decoded segment checked for plausible duration, peak and characters-per-second.
- [x] Stopped an unbatched conversion and resumed it with batching from its SQLite watermark — no sentences redone outside the interrupted block.
- [ ] Full-length batched conversion to final `.m4b`: a 10,153-sentence book is mid-run at time of writing.


lib/classes/device_installer.py
```diff
@@ -1772,7 +1771,9 @@ def _needs_reinstall():
                 return True
             m_ta = re.search(r'\+(.+)$', torchaudio_full)
             torchaudio_tag = m_ta.group(1) if m_ta else None
-            if not _tag_ok(torchaudio_tag):
+            # Unlike torch, torchaudio wheels don't always carry a device-local version
+            # tag; only enforce the match when torchaudio actually reports one.
+            if torchaudio_tag is not None and not _tag_ok(torchaudio_tag):
```
aaron-ang, line 1776: Treating a bare torchaudio version (no device-local tag) as a mismatch previously caused every run to force a reinstall, even when torch/torchaudio were already correct.

ext/py/breeze-tts/breeze_infer/api.py
```diff
@@ -0,0 +1,478 @@
...
+    *,
+    instruction: str,
+    cfg_scale: float,
+    seed: int,
+    reference_path: Path | None,
+    ref_text: str,
+) -> list[bytes]:
+    """Synthesize many texts in one batched forward pass.
+
+    Bypasses FastBreezeStreamingRuntime, whose CUDA-graph batch dimension is
```
aaron-ang, line 288: Why the batch endpoint doesn't reuse the fast path: `FastBreezeStreamingRuntime`'s CUDA-graph batch dimension is consumed entirely by CFG (cond + uncond), so it cannot batch texts at all. Eager `generate()` is slower for a single sequence but accepts a real batch.

The win comes from decode being memory-bandwidth bound rather than compute bound: the backbone and the depth decoder re-read their full weights for every audio frame, so at batch 1 throughput is capped far below what the GPU can compute. Batching amortizes that weight read across sentences.

ext/py/breeze-tts/breeze_infer/api.py
```diff
@@ -0,0 +1,478 @@
...
+        app.state.audio_tokenizer,
+        app.state.model,
+        requests,
+        get_template(template_name),
+        guidance_scale=cfg_scale,
+        guidance_scale_ref=None,
+        guidance_scale_ins=None,
+    )
+
+    # the fast path applies REPETITION_PENALTY through FastStreamingConfig; eager
```
aaron-ang, line 320: Parity detail: the fast path applies `REPETITION_PENALTY` through `FastStreamingConfig`, so eager `generate()` has to be handed the processor explicitly or the two paths would sample differently.

ext/py/breeze-tts/breeze_infer/api.py
```diff
@@ -0,0 +1,478 @@
...
+REPO_ROOT = Path(__file__).resolve().parents[1]
+FAST_CONFIG = REPO_ROOT / "configs" / "fast.json"
+DEFAULT_CFG_SCALE = 1.0
+MAX_NEW_TOKENS = 1500
+MAX_SEQ_LEN = 2048
+REPETITION_PENALTY = 1.1
+OPTIONAL_AUDIO_FILE = File(None)
+# Once decode stops being bandwidth bound, larger batches stop paying for
+# themselves while padding waste keeps growing.
+MAX_BATCH_TEXTS = 128
```
aaron-ang, line 50: The cap is a judgement call, not a hardware limit. Past the point where decode stops being bandwidth bound, larger batches stop paying for themselves while padding waste keeps growing — a batch runs until its longest sequence finishes.

ext/py/breeze-tts/breeze_infer/api.py
```diff
@@ -0,0 +1,478 @@
...
+    finally:
+        if reference_path is not None:
+            reference_path.unlink(missing_ok=True)
+        _request_lock.release()
+
+    return Response(
+        content=b"".join(segments),
+        media_type="application/octet-stream",
+        headers={
+            "X-Segment-Bytes": ",".join(str(len(segment)) for segment in segments),
```
aaron-ang, line 426: Wire format: the body is the concatenated s16le PCM of every segment in request order, and `X-Segment-Bytes` gives each segment's length so the caller can split the blob apart. One binary body rather than a JSON array of base64 strings, which would inflate the audio by a third.

ext/py/breeze-tts/breeze_infer/api.py
```diff
@@ -0,0 +1,478 @@
...
+                status_code=400,
+                detail="ref_audio and ref_text must be provided together or both omitted.",
+            )
+        if has_reference:
+            assert ref_audio is not None
+            reference_path = await _save_upload(ref_audio)
+
+        # generation is synchronous and CPU/GPU bound; keep it off the event loop
+        # so /health stays responsive for the whole minutes-long batch
+        segments = await run_in_threadpool(
```
aaron-ang, line 408: Generation is synchronous and GPU-bound for minutes at a time. Running it in the threadpool keeps the event loop free so `/health` still answers while a batch is in flight — the engine polls `/health` to decide whether to spawn a server.

lib/classes/tts_engines/breeze.py
```diff
@@ -0,0 +1,388 @@
...
+                    "breeze_infer.api",
+                    weights_dir,
+                    "--host",
+                    BREEZE_API_HOST,
+                    "--port",
+                    str(BREEZE_API_PORT),
+                ]
+                # --fast-all's CUDA-graph batch dimension is consumed by CFG, so the
+                # batch endpoint cannot use it and the warmup would be wasted.
+                if self.batch_size == 1:
```
aaron-ang, line 159: `--fast-all` is skipped whenever batching is on. Its captured CUDA graphs cannot serve the batch endpoint (see the comment in `api.py`), so capturing them is pure startup cost — time-to-healthy dropped from roughly 105 s to 70 s on the test box.

lib/classes/tts_engines/breeze.py
```diff
@@ -0,0 +1,388 @@
...
+                # --fast-all's compile path needs Triton's ptxas to support the GPU
+                # target; the bundled torch/triton one may not on newer architectures.
+                env = os.environ.copy()
+                # Respect an already-set TRITON_PTXAS_PATH; only fall back to the
+                # system CUDA toolkit's ptxas.
+                env.setdefault("TRITON_PTXAS_PATH", "/usr/local/cuda/bin/ptxas")
+                # sys.executable, not "python": PATH may point at an interpreter
+                # without breeze_infer installed.
+                command = [
+                    sys.executable,
```
aaron-ang, line 148: Bug fix: this was a bare `"python"`, resolved through `PATH`. It only worked when the launching shell happened to have this project's venv active; from anywhere else it picked up an unrelated interpreter and died with `No module named 'breeze_infer'`.

lib/classes/tts_engines/breeze.py
```diff
@@ -0,0 +1,388 @@
...
+        except (KeyError, OSError, RuntimeError, ValueError) as e:
+            # load_engine() already wraps its own failures; anything outside this set
+            # is a bug here and should keep its traceback.
+            error = f"__init__() error: {e}"
+            raise ValueError(error) from e
+
+    # tells lib.core it can call convert_batch() instead of one convert() per sentence
+    supports_batching = True
+
+    def _resolve_batch_size(self) -> int:
```
aaron-ang, line 50: Batch size comes from the `breeze` preset (64) and can be retuned per GPU with `E2A_BREEZE_BATCH_SIZE`. Non-integer or `< 1` values are reported and ignored rather than crashing a long conversion at startup.

lib/classes/tts_engines/breeze.py
```diff
@@ -0,0 +1,388 @@
...
+                segments.append(
+                    np.frombuffer(raw, dtype="<i2").astype(np.float32) / 32768.0
+                )
+        return segments
+
+    def convert_batch(self, items: list, **kwargs) -> tuple:
+        """Synthesize many sentences per inference request.
+
+        ``items`` is a list of ``(sentence_file, sentence)``. A sentence can mix
+        speech with SML tags, so parts are flattened into one flat list of texts
```
aaron-ang, line 326: A sentence can mix speech with SML tags, so parts are flattened into one list of texts for the model and reassembled per sentence afterwards. SML tags are handled locally by `_convert_sml()` and never reach the server, so they don't occupy a batch slot.

lib/classes/tts_engines/breeze.py
```diff
@@ -0,0 +1,388 @@
...
+    def _server_is_up(self) -> bool:
+        # /health returns 503 while the model is still loading, 200 once ready.
+        import requests
+
+        try:
+            r = requests.get(
+                f"http://{BREEZE_API_HOST}:{BREEZE_API_PORT}/health", timeout=2
+            )
+            return r.status_code == 200
+        except requests.RequestException:
```
aaron-ang, line 78: Ruff BLE001: every blind `except Exception` in this file is narrowed to what the guarded code can legitimately raise, so real bugs stop being swallowed alongside the failures these handlers exist to absorb. Here specifically, only a network failure means "not up yet" — a bug in this method must not masquerade as an unhealthy server and silently time out the wait loop.

`convert()`'s imports had to be hoisted above its `try` so the `except` clause can name `requests.RequestException` without a `NameError`.

lib/core.py
```diff
@@ -2921,12 +2952,46 @@ def _count_sentences(sentences:list)->int:
...
+                    pending = [j for j in sorted(valid_idx) if j >= start_sentence or j in missing_sentences]
+                    skipped = len(valid_idx) - len(pending)
+                    if skipped > 0:
+                        t.update(skipped)
+                    if pending and start_sentence > 0:
+                        show_alert(session_id, {'type': 'info', 'msg': f'*** Resuming from sentence {global_sent + skipped} ***'})
+                    # A batch runs until its longest sequence finishes, so mixing lengths
+                    # wastes compute; character count tracks duration closely enough to
+                    # group similar sentences together.
+                    pending.sort(key=lambda idx: len(sentences[idx].strip()))
```
aaron-ang: Scheduling: a batch runs until its longest sequence finishes, so mixing lengths wastes compute on padding. Character count tracks spoken duration closely enough to group similar sentences together, which lifts occupancy well above source order.

lib/core.py
```diff
@@ -2921,12 +2952,46 @@ def _count_sentences(sentences:list)->int:
...
+                        items = [
+                            (os.path.join(block_dir, f'{j}.{default_audio_proc_format}'), sentences[j].strip())
+                            for j in group
+                        ]
+                        run, error = tts_manager.convert_sentences2audio(items, block_voice=block_voice)
+                        if not run:
+                            show_alert(session_id, {'type': 'warning', 'msg': error})
+                            return False
+                        converted = True
+                        # sentence_resume is not advanced mid-block: batched sentences
```
aaron-ang: Resume semantics, deliberately different from the sequential path: `sentence_resume` is not advanced mid-block here. Batched sentences finish out of order, so a single watermark cannot describe partial progress. An interrupted block is redone whole, and the sentence files already inside it are reused rather than regenerated.

lib/core.py
```diff
@@ -2804,6 +2804,37 @@ def _check_block_sentences(block_id:str, sentences:list)->set:
     def _count_sentences(sentences:list)->int:
         return sum(1 for s in sentences if any(c.isalnum() for c in s.strip()))
 
+    def _mark_baseline()->None:
```
aaron-ang: These three helpers exist because the batched and sequential paths need the same baseline-init, progress-update and end-of-block-combine steps. Before the batch path was added there was one copy; factoring them out avoided a second.

aaron-ang (commented):
Inline notes on the rationale behind the less obvious choices, kept out of the PR description.

lib/core.py
```diff
@@ -2804,6 +2804,37 @@ def _check_block_sentences(block_id:str, sentences:list)->set:
...
+        baseline_initialized = True
+
+    def _advance_progress(count:int, desc:str)->None:
+        t.update(count)
+        total_progress = t.n / total_sentences
+        if session['is_gui_process']:
+            progress_bar(progress=total_progress, desc=f'{ebook_name} - {desc}')
+        t.set_description(f'{total_progress * 100:.2f}%')
+
+    def _finalize_block(ch_num:int, x:int, block_id:str, chapter_audio_file:str, block_len:int, sent_start:int, sent_end:int, combine:bool)->bool:
```
aaron-ang: `combine` is the old `if converted or block_changed or missing_sentences` condition, passed in rather than recomputed, so both paths share one definition of "this block needs recombining".

lib/core.py
```diff
@@ -3773,7 +3826,8 @@ def convert_ebook(args:dict)->tuple:
                 session['abs_api_token'] = str(args.get('abs_api_token', ''))
                 session['abs_library'] = str(args.get('abs_library', ''))
                 os.makedirs(session['voice_dir'], exist_ok=True)
-                audio_pre_final_exist = os.path.exists(os.path.join(session['process_dir'], ebook_name + '.' + default_audio_proc_format))
+                audio_pre_final_file = os.path.join(session['process_dir'], ebook_name + '.' + default_audio_proc_format)
```
aaron-ang, line 3790: Unrelated pre-existing bug, found while resuming a conversion. See the next comment.

lib/core.py
```diff
@@ -3790,7 +3844,7 @@ def convert_ebook(args:dict)->tuple:
                         print("Please enter 's', 'r' or 'd'.")
                     if choice == 'r':
                         if audio_pre_final_exist:
-                            os.unlink(audio_pre_final_exist)
+                            os.unlink(audio_pre_final_file)
```
aaron-ang, line 3808: This was `os.unlink(audio_pre_final_exist)` — the bool from `os.path.exists()`, not the path — so choosing `[r]esume` raised `TypeError` whenever a pre-final audio file actually existed. Holding the path in its own variable fixes it.

ROBERT-MCDOWELL:
before to PR, provide the version of e2a.
1) provide the breeze-tts repo to check what it is and what license is used
2) file architecture must be respected, meaning breeze-tts.py must have all owned functions.
3) requirements.txt must not be touched.
4) audio_pre_final_file fixed
5) def_finalize_block() patch makes no sense at all. explain more where is the issue.

aaron-ang:
Version: `26.8.20`, branch rebased on `main` @ `593a5b2c` (0 behind). Pushed `10660d04`.

**1) breeze-tts repo and license**

Upstream code: https://github.com/breezeblue-ai/breeze-tts — Apache-2.0, kept verbatim in `ext/py/breeze-tts/LICENSE`. Its audio tokenizer derives from Qwen3-TTS, also Apache-2.0.

Weights are **not** Apache: https://huggingface.co/BreezeBlue/breeze-tts-2 ships under the BreezeBlue Research and Non-Commercial License, which grants no commercial rights. That is stated next to the engine in the README and in the header of `ext/py/breeze-tts/setup.py`. The only edit to upstream source is renaming its `models` package to `breeze_models`, which otherwise collides on install.

**2) file architecture — breeze-tts.py owns its own functions**

Fixed. `supports_batching`/`batch_size` are gone from `TTSManager`, and `core.py` has no engine-specific branch left. `breeze.py` owns both new methods:

- `Breeze.plan_chunks()` — groups sentence indices, length-sorted
- `Breeze.convert_batch()` — one inference request per group

`TTSManager` only dispatches, with defaults that make every other engine byte-identical to before: no `plan_chunks` means one sentence per chunk, no `convert_batch` means the existing per-sentence `convert()` loop.

**3) requirements.txt must not be touched**

Fixed. The diff is now one line, `./ext/py/breeze-tts`, next to the existing `./ext/py/demucs`. `qwen-tts`, `uvicorn` and `python-multipart` moved into the vendored package's own `setup.py`. torch/torchaudio/transformers are deliberately not pinned there so `DeviceInstaller` keeps resolving them per device.

**4) audio_pre_final_file**

This one is not fixed on `main`. At `lib/core.py:3793`:

```python
audio_pre_final_exist = os.path.exists(...)   # line 3776, a bool
...
if audio_pre_final_exist:
    os.unlink(audio_pre_final_exist)          # line 3793, unlink(True)
```

`os.unlink(True)` raises `TypeError` on any resume where the user answers `r` and a pre-final file exists. The patch keeps the path in `audio_pre_final_file` and unlinks that. Happy to split it into its own PR if you'd rather not carry it here.

**5) the _finalize_block() patch**

You're right, and it's removed. It was a wrapper I extracted to avoid duplicating the block-combine tail across two sentence loops — which was the wrong fix, because the two loops shouldn't have existed. `convert_chapters2audio()` is now a single chunked loop, so there's nothing to deduplicate and the helper is gone.

The one real change left in that function: `sentence_resume` used to be `= j` inside a strictly ascending loop. Batched chunks can complete out of order, so a single index can't describe progress with a hole behind it. It now advances only across a contiguous run of completed sentences, which is bit-identical to the old behaviour for every one-sentence-per-call engine.

ROBERT-MCDOWELL:
I don't need A.I. comments, I need yours.

aaron-ang:
hi @ROBERT-MCDOWELL here's my take: I stand by everything I posted (AI or not).

Regarding the unlink bug: I still can't find the fix in v26 branch; the one that you recently merged to.

I really like your project as a base to create audiobooks, but it is hard to understand how the project runs. There are many adhoc files and scripts scattered around the project. Furthermore, there is no project guidelines (such as CONTRIBUTING.md) that makes it hard for contributors to understand the structure of the project and inherent assumptions, so I did my best to carve out the changes based on similar files, fixing pre-existing bugs along the way.

I decided to create this PR because Breeze TTS-2 is now the top open weights model and I decided to self-host to create my own audiobooks. Either way, the code is running live on my NVIDIA GB10 box, and the decision to merge my changes to your project is yours to make.

Edit: just saw your fix commits an hour ago. you didn't have to close PRs. i can rebase on top of your changes.

ROBERT-MCDOWELL:
it's always better to talk human to human rather than copy paste A.I. text and code which is at 95% not the final solution since they lack of intuition and real intelligence of the project architecture. More, they always want to impose and add new files and extend the architecture which at the end is a kind of spaghettis code on a castel card that only A.I. will be able to patch until the whole thing will collapse and no human will pass their life to debug.
This rhetoric said, Drew and I decided to not add more TTS on the list with the actual architecture as the requirements deps pkg versions became impossible to manage between them. We are working on a new dynamic modular architecture that will unblock the constraint.
concerning breeze-tts, we will add it once it will be also compatible with CPU.

aaron-ang:
@ROBERT-MCDOWELL I would argue most human code produces less than 95% completeness (let alone the final solution). Obviously the maintainer knows the code the best as they oversee all code changes, but it is also important for the project to be structured well so that external contributors can extend the project easily. This is something AI is good at: inferring patterns from code. Sure, it probably won't get things right the first try, but based on my experience working on other open source projects, well-documented code which also follows coding best principles always yields the best outcomes for both the maintainers and contributors (whether using AI or not).

It's good to know that the team is looking to make broader changes. Looking forward to the new dynamic modular architecture and more models to be added.

ROBERT-MCDOWELL:
A.I. (depend which model) is ok to code, not to maintain a repo. If you let the A.I. to manage your repo you will finish in a massive spaghettis code, especially when you have to trick some parts to make it work which A.I. does not understand at all unless you pass your time to train it. an A.I. will never has what is called "intuittion" that solves many issues without to create mountains of code.

PR #2095, opened 2026-09-18
fix: align CUDA 13.2 Torch packages and XTTS runtime

Joghurt:
- install matching Torch, Torchaudio, and Torchvision Nightly builds
- add UniDic Lite compatibility for XTTS
- ignore generated voices and device metadata
- support Windows CUDA 13.2 tags


ROBERT-MCDOWELL:
your patch mixes unrelated and nonssense code. check conf.py, cuda 13.2 is already managed.

